PR #15, opened 2019-11-28
Fix templates

derom:
With current templates ```helm template``` produces incorrect manifest, like so:
```
---kind: ConfigMap
apiVersion: v1
```
Also, ```limit``` is invalid field


khaines:
#25 covered these items, so I'm going to close this PR. Thank you again

PR #25, opened 2020-04-25
refactor for cortex v1.1.0

previ:
A major refactoring of the initial helm chart.
- It installs cortex in "microservices" mode, using the single docker image but with specific "targets".
- It also leverages one single configmap to store the configuration as a cortex.yaml file mounted by all components.
-  It also (optionally) installs both memcached as cache and postgres as db for configsdb.
- the mandatory kv store is not provided, as it is considered an external dependency.

I have been testing and using the chart with cortex 1.0.0 on Azure AKS, cassandra and azure blob as persistent storage. No issues so far (a few weeks of continuous use). Not tested with aws (dynamodb/s3) but the. configuration is there, commented, it should work.


hairyhenderson:
@previ - it looks like you need to sign-off your commits to allow the DCO bot to pass (see https://github.com/cortexproject/cortex-helm-chart/pull/25/checks?check_run_id=633880585 for details on how).

/cc @gramidt - any chance on getting this reviewed?

previ:
Signed off, thanks @hairyhenderson.

templates/configs-dep.yaml
```diff
@@ -1,49 +1,110 @@
...
+      tolerations:
+        {{- toYaml .Values.configs.tolerations | nindent 8 }}
+      terminationGracePeriodSeconds: {{ .Values.configs.terminationGracePeriodSeconds }}
+      volumes:
+        - name: config
+          secret:
+            secretName: {{ template "cortex.fullname" . }}
+        - name: postgres-password
+          secret:
+            secretName: {{ template "cortex.fullname" . }}-configsdb-postgresql
```
xeor, line 104: I can't find anywhere this secret is defined. Should be made manually?

previ: It's actually created by the postgresql chart.

Chart.lock
```diff
@@ -0,0 +1,15 @@
...
+- name: memcached
+  repository: https://kubernetes-charts.storage.googleapis.com/
+  version: 3.2.3
+- name: memcached
+  repository: https://kubernetes-charts.storage.googleapis.com/
+  version: 3.2.3
+- name: memcached
+  repository: https://kubernetes-charts.storage.googleapis.com/
+  version: 3.2.3
+- name: postgresql
```
khaines, line 11: While the memcache dependencies were here in the previous chart, I would insist we not add any more dependencies to the chart. There are good cases for not running/managing postgresql (or memcache) within one's k8s cluster, so we shouldn't require it to be installed by the cortex chart. This changes a few of your settings and value assumptions below if removed.

previ: I can definitely modify the chart in order to make postgresql as well as memcache optional. Memcache should be optional already, if it's not, it's a bug. Need to check it out.
I'll come back soon with an updated chart.

previ: I just committed a few updates which makes both postgresql and memcache optional.
Posgresql gets replaced by "memory://".
Default values.yaml disable both postgresql and memchace (all instances).

khaines: No, that's not what I was intending with my comment. Your latest change only makes using the chart deployed postgresl optional, with no option to use an externally created instance. This is why I think it's a better pattern to not assume postgresql as a chart dependency here and just require the needed settings to communicate to an instance of it in the values config.

previ: Ah ok, I got it now. In my use case it's convenient to have the chart to be almost fully self contained (let apart consul and cassandra) but I also understand most people will use an existing/external postgresql instance.

previ: question: as postgresql password must be kept in secret (mounted as password file), which solution is preferable:
1. depend on an externally created secret.
2. create a secret, specific for postgresql, with the password (as read from values).
Personally I prefer 1 as it already fits with an helm based installation of postgresql (which creates the secret).

previ: Chart updated: secret with postgresql pwd can be used if available, or created with specified pwd.
Also image tag updated to recently released v1.1.0.

Chart.yaml
```diff
@@ -1,4 +1,30 @@
-apiVersion: v1
-description: Chart for installing cortex.
+apiVersion: v2
+version: 0.1.2
+appVersion: v1.0.0
```
khaines, line 3: should this be v1.1.0 now?

templates/configs-dep.yaml
```diff
@@ -1,49 +1,118 @@
...
-        - -server.register-instrumentation=true
-        {{ if .Values.configsdb.enabled -}}
-        - -database.uri=postgres://postgres@{{ template "fullname" . }}-configs-db.{{ .Release.Namespace }}.svc.cluster.local/configs?sslmode=disable
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args:
+            - "-target=configs"
+            - "-config.file=/etc/cortex/cortex.yaml"
+            {{- if .Values.configsdb_postgresql.enabled }}
```
khaines, line 57: this section still seems to read like it only uses postgresql if it is enabled, meaning using the chart dependency installed version.

khaines: had to read closer, but this looks good.

Chart.yaml
previ: Correct, I just pushed an update to v1.1.0.

DandyDeveloper:
@previ I'm just using this to test a POC, one thing that may be worth mentioning in the README is that both Consul as an external dependencies no longer included. It's removed in your PR.

DandyDeveloper:
@previ @khaines https://github.com/cortexproject/cortex-helm-chart/pull/28

Just an FYI as well, it may be worth including this or letting me know when you've merged the above as it'll likely cause a conflict.

khaines:
@DandyDeveloper I'm going to let this age a bit for any more comments and merge it before weeks end. It will cause a conflict in your PR certainly.

previ:
> @previ I'm just using this to test a POC, one thing that may be worth mentioning in the README is that both Consul as an external dependencies no longer included. It's removed in your PR.

Yes I can update the README and document the external dependencies, which are: consul, storage (choice of cassandra, dynamodb, tsdb).
I guess consul was not provided as dependency in the original chart, but it's definitely required.

DandyDeveloper:
@khaines We ready to merge this?

drequena:
My 2 cents to the Chart.

First, congrats, the Chart is awesome and I'm using as the base for a large Cortex setup.

IMHO, just like memcached and configsdb. Ruler and Alertmanager should have a "enable" property since they are optional in the Cortex Architecture.

I noticed that the default values of the Chart are all set to run a setup as simple as possible. That's fine, but this decision should be documented somehow. It took me a long time to notice parameters like: join_after and max_transfer_retries (ingester->lifecycler session) were the cause to my deployment rollout get stuck every time.

The nginx component is not necessary if you have a ingress controller installed on your cluster. A conditional could be placed on ingress object template, that if nginx is disable and ingress is enable, the routing rules would be placed inside ingress object.

My final point. The decision of having a single config file (secret) for all components can be tricky sometimes. A single parameter to the distributor triggered a complete rollout of all components (helm upgrade --install ... etc.)

previ:
> My 2 cents to the Chart.
> 
> First, congrats, the Chart is awesome and I'm using as the base for a large Cortex setup.
Thanks! :-)

> IMHO, just like memcached and configsdb. Ruler and Alertmanager should have a "enable" property since they are optional in the Cortex Architecture.

Yes, I agree. Actually memcache is optional (it can be disabled). Ruler and Alertmanager are not optional, I will work on this, should be easy.

> I noticed that the default values of the Chart are all set to run a setup as simple as possible. That's fine, but this decision should be documented somehow. It took me a long time to notice parameters like: join_after and max_transfer_retries (ingester->lifecycler session) were the cause to my deployment rollout get stuck every time.

My bad, I do not have the required expertise on cortex to be able to provide a more tuned configuration. Any help is appreciated.

> The nginx component is not necessary if you have a ingress controller installed on your cluster. A conditional could be placed on ingress object template, that if nginx is disable and ingress is enable, the routing rules would be placed inside ingress object.

Good point, I'll work on that.
 
> My final point. The decision of having a single config file (secret) for all components can be tricky sometimes. A single parameter to the distributor triggered a complete rollout of all components (helm upgrade --install ... etc.)

This is possibly more an approach on the how to manage configuration: I personally prefer a single config file as a "source of truth", even if this implies unnecessary restarts.
As far as I know, it's not possible to have a config file for common parameters and separate files for specific parameters. But again, I might have missed something.

xeor:
@drequena : Du you mind share what configuration you figured works for creating a more production-grade cluster?

https://github.com/cortexproject/cortex/issues/2424 is about a "production-grade helm chart" and was worked on by @Konpon96. Maybe he has any comments as well?

xeor:
Another thing I found was, the doc recommends using statefulsets (https://cortexmetrics.io/docs/production/ingesters-with-wal/#migrating-from-stateless-deployments) for the ingesters in production. They are now normal deployments. Should this be configurable or changed?

previ:
> Another thing I found was, the doc recommends using statefulsets (https://cortexmetrics.io/docs/production/ingesters-with-wal/#migrating-from-stateless-deployments) for the ingesters in production. They are now normal deployments. Should this be configurable or changed?

Yes, statefulsets are required to take advantage of WAL.

About other aspects of the chart to be more "production grade": in my opinion the chart should be flexible, so provide a basic install as a default, as well configurable for more production deployments.
Of course if some default config settings in the chart values are just wrong, they must be fixed.

kisahm:
+1

xeor:
@previ Do you see any negative effects of having statefulsets instead as the default?

It is not something that can be changed without patching the whole chart, so not having that as default (or a configuration option) makes it harder to use for production, since you need to patch the chart anyway.

previ:
@xeor Yes I am working on it. The idea is to move the ingester to statefulset but leave the PVC (needed by WAL) optional.

khaines:
I was hoping to get a 2nd set of maintainer eyes here, but I don't want this to just keep growing either, so let's get it in and iterate on it

Issue #27, opened 2020-05-27
Nginx Component assumes DNS Resolver is kube-dns


DandyDeveloper:
The nginx config doesn't allow an override of the DNS resolver you would prefer it to use.

https://github.com/cortexproject/cortex-helm-chart/blob/master/templates/nginx-config.yaml#L32

In our case, we're using CoreDNS. It would be nice to be able to change this. I can make a quick PR.

PR #28, opened 2020-05-27
Adding dnsResolver option to nginx.config

DandyDeveloper:
Signed-off-by: Aaron Layfield <aaron.layfield@gmail.com>

fixes: #27


khaines:
@DandyDeveloper, can you rebase with master for the latest changes from #25? if this PR is still needed post that rebase, I'll merge it in

DandyDeveloper:
@khaines All done. I think it's still worth having, as in our case we use kube-proxy, and by default, this chart would only manage kube-dns based CNI

Issue #29, opened 2020-06-30
Add enable flag for the Nginx proxy


rsteneteg:
In our setup we don't need the nginx component since we have our own proxy components in front of Cortex.

PR #32, opened 2020-07-02
Making Nginx optional

khaines:
Instances of cortex that have their own frontdoor/proxy can turn off nginx now.

Addresses #29 

Signed-off-by: Ken Haines <ken@kenhaines.net>


rsteneteg:
Hi Ken, thanks for making this change

Since the ingress resource is using the `.Values.nginx.service.port` it makes sense to also disable the ingress resource if nginx is disabled. I added this in my local copy of ingress,yaml

`{{- if and .Values.ingress.enabled .Values.nginx.enabled -}}`

khaines:
Great callout Roger. I've updated the if statement

rsteneteg:
Thanks Ken, LGTM

Issue #33, opened 2020-07-04
Queriers are not able to connect to the Query frontend


rsteneteg:
After configuring the queriers to connect to the frontend using this config

```
frontend_worker:
      frontend_address: cortex-query-frontend-headless.<namespace>.svc.cluster.local:9095
```

The queriers are not connecting and the query frontend gets into a crashloop because no queriers has connected

`level=info ts=2020-07-03T14:55:35.497218784Z caller=frontend.go:454 msg="not ready: number of queriers connected to query-frontend is 0"`

The queriers are not logging any errors or information about why it is not connecting.

PR #34, opened 2020-07-04
Cortex uses DNS SRV records and looks for grpclb as the service name.…

rsteneteg:
Cortex uses DNS SRV records and looks for `grpclb` as the service name. Publishing not ready addresses is also required since the frontends wont go ready until a querier has connected, and the queriers wont find any frontend SRV records until they are ready

Addresses #33 

Signed-off-by: Roger Steneteg <rsteneteg@ea.com>


khaines:
Thank you Roger! Simple correction, but a much-needed one.

Issue #45, opened 2020-08-31
Add support to run Ingesters as a StatefulSet


rsteneteg:
In order to better support running Cortex with WAL the option to run the ingesters as a statefulset is preferred.

PR #46, opened 2020-09-03
adding option to deploy ingester as statefulset

rsteneteg:
Signed-off-by: Roger Steneteg <rsteneteg@ea.com>

Fixes: #45


rsteneteg:
@khaines all components have a `<component>.persistence.subpath` key, for the PVC I added a `ingester.persistentVolume` section since "persistentVolume" is what prometheus uses for the PVC and I think it's more common. I marked the "ingester.persistence" as deprecated, thoughts?

khaines (approved):

PR #48, opened 2020-09-11
Fix target labels

thibautcharry:
The target label is missing on several deployments and it is wrong on the querier deployment.
This simple PR is to have a consistent and correct target label on all deployed components.

Signed-off-by: Thibaut Charry <ttcharry@gmail.com>


khaines (approved):
Thank you for the submission!

PR #50, opened 2020-09-13
Blocks storage support in Cortex Helm chart

### LINKED ISSUE #49: Support Blocks storage
Currently, the helm chart only supports the Chunks storage mode for cortex. The helm chart should also support configurations and components for the use of blocks storage.

khaines:
Add extra components and options for utilizing the newer blocks storage mode of Cortex.

Resolves #49

### DIFF
--- a/templates/compactor-statefulset.yaml
+++ b/templates/compactor-statefulset.yaml
@@ -1,55 +1,55 @@
-{{- if .Values.ingester.statefulSet.enabled -}}
+{{- if and .Values.compactor.enabled | eq .Values.config.storage.engine "blocks" -}}
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
-  name: {{ template "cortex.fullname" . }}-ingester
+  name: {{ template "cortex.fullname" . }}-compactor
   namespace: {{ .Release.Namespace }}
   labels:
-    app: {{ template "cortex.name" . }}-ingester
+    app: {{ template "cortex.name" . }}-compactor
     chart: {{ template "cortex.chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
   annotations:
-    {{- toYaml .Values.ingester.annotations | nindent 4 }}
+    {{- toYaml .Values.compactor.annotations | nindent 4 }}
 spec:
-  replicas: {{ .Values.ingester.replicas }}
+  replicas: {{ .Values.compactor.replicas }}
   selector:
     matchLabels:
-      app: {{ template "cortex.name" . }}-ingester
+      app: {{ template "cortex.name" . }}-compactor
       release: {{ .Release.Name }}
   updateStrategy:
-    {{- toYaml .Values.ingester.statefulStrategy | nindent 4 }}          
-  serviceName: {{ template "cortex.fullname" . }}-ingester
-  {{- if .Values.ingester.persistentVolume.enabled }}
+    {{- toYaml .Values.compactor.strategy | nindent 4 }}            
+  serviceName: {{ template "cortex.fullname" . }}-compactor
+  {{- if .Values.compactor.persistentVolume.enabled }}
   volumeClaimTemplates:
     - metadata:
         name: storage
-        {{- if .Values.ingester.persistentVolume.annotations }}
+        {{- if .Values.compactor.persistentVolume.annotations }}
         annotations:
-          {{ toYaml .Values.ingester.persistentVolume.annotations | nindent 10 }}
+          {{ toYaml .Values.compactor.persistentVolume.annotations | nindent 10 }}
         {{- end }}
       spec:
-        {{- if .Values.ingester.persistentVolume.storageClass }}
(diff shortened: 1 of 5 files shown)


values.yaml
```diff
@@ -882,6 +893,240 @@ nginx:
...
+
+  resources:
+    limits:
+      cpu: 1
+      memory: 1Gi
+    requests:
+      cpu: 100m
+      memory: 512Mi
+
+  ## Additional Loki container arguments, e.g. log level (debug, info, warn, error)
```
rsteneteg, line 912: "Loki" arguments :)

values.yaml
```diff
@@ -882,6 +893,240 @@ nginx:
...
+                - key: target
+                  operator: In
+                  values:
+                    - store-gateway
+            topologyKey: 'kubernetes.io/hostname'
+
+  annotations: {}
+
+  ## DEPRECATED: use persistentVolume.subPath instead
+  persistence:
```
rsteneteg, line 942: Since this is a new component that did not exist before.. could be remove the persistence.subpath option and only do persistentVolume ?

templates/ingester-statefulset.yaml
```diff
@@ -17,6 +17,8 @@ spec:
     matchLabels:
       app: {{ template "cortex.name" . }}-ingester
       release: {{ .Release.Name }}
+  updateStrategy:
+    {{- toYaml .Values.ingester.statefulStrategy | nindent 4 }}
```
rsteneteg, line 21: do we need a different statefulStrategy value here compared to using the strategy value the other components use. With the addition of the new statefulset components they use strategy so currently only ingester would use statefulStrategy

values.yaml
```diff
@@ -882,6 +893,240 @@ nginx:
...
+                - key: target
+                  operator: In
+                  values:
+                    - compactor
+            topologyKey: 'kubernetes.io/hostname'
+
+  annotations: {}
+
+  ## DEPRECATED: use persistentVolume.subPath instead
+  persistence:
```
rsteneteg, line 1059: same question here as with the store-gateway if we can just remove this since it is a new component

values.yaml
```diff
@@ -882,6 +893,240 @@ nginx:
...
+
+  resources:
+    limits:
+      cpu: 1
+      memory: 1Gi
+    requests:
+      cpu: 100m
+      memory: 512Mi
+
+  ## Additional Loki container arguments, e.g. log level (debug, info, warn, error)
```
rsteneteg, line 1029: "Loki" args again :)

templates/ingester-statefulset.yaml
khaines: Yes the update strategy structure is different between deployments and statefulsets, so ingester ends up with 2 settings :( because it is the only component here that can be a deployment or statefulset

rsteneteg: ok, valid point, thanks

values.yaml
khaines: yeah, good call out. they were in whatever source file I copied the template from. I cleaned all references up in the values file, since they were there from the large refactor.

values.yaml
khaines: Yes, let's do that

values.yaml
khaines: done

templates/store-gateway-statefulset.yaml
```diff
@@ -0,0 +1,147 @@
...
+        resources:
+          requests:
+            storage: "{{ .Values.store_gateway.persistentVolume.size }}"
+  {{- end }}
+  template:
+    metadata:
+      labels:
+        app: {{ template "cortex.name" . }}-store-gateway
+        name: {{ template "cortex.name" . }}-store-gateway
+        target: store_gateway
```
thibautcharry, line 50: With regards to #48 where `target: table-manager` was used, I would use a `-` instead of `_` if you don't see any issue with this.

templates/compactor-statefulset.yaml
```diff
@@ -0,0 +1,147 @@
+{{- if eq .Values.storage.engine "blocks" -}}
```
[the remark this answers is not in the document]
thibautcharry: The `storage.engine` value does not exist in the values file. Was the intent to use `.Values.config.storage.engine`?

templates/store-gateway-statefulset.yaml
khaines: Great callout. I did intend for this to be a `-` vs `_`

templates/compactor-statefulset.yaml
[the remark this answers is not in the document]
khaines: Yep already fixed the values pathing. However, as much as the compactor is optional in the docs in practicality you need the compactor in any deployments that are not a "single-app" developer instance. I don't mind adding an enabled flag for this, but I will set the default to `true`.

rogst:
Thank you Ken, LGTM

PR #51, opened 2020-09-15
Fix values file duplicate config.storage section

thibautcharry:
In #50, the `config.storage` section in the values.yaml file has been duplicated by mistake.
This causes `helm lint` to fail:
```
helm lint .
==> Linting .
[ERROR] templates/: template: cortex/templates/table-manager-svc.yaml:1:7: executing "cortex/templates/table-manager-svc.yaml" at <ne .Values.config.storage.engine "blocks">: error calling ne: incompatible types for comparison

Error: 1 chart(s) linted, 1 chart(s) failed
```
The `engine` key is moved to the already existing `config.storage` section.

Signed-off-by: Thibaut Charry <ttcharry@gmail.com>

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -57,9 +57,6 @@ config:
     enforce_metric_name: false
     reject_old_samples: true
     reject_old_samples_max_age: 168h
-    # See https://github.com/cortexproject/cortex/blob/master/docs/configuration/config-file-reference.md#storage_config
-  storage:
-    engine: chunks
   schema:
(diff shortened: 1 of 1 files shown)


thibautcharry:
@khaines Small error that was introduced in your latest changes 🙂

khaines (approved):
Interesting, great catch! @thibautcharry. I'll have to see why the build test for this passed.

PR #52, opened 2020-09-22
add grpc ports to store gateway pods and service

khaines:
The store gateway utilizes GRPC for data access, however, these ports are missing from the template YAML. This PR corrects that.


rsteneteg:
LGTM

Issue #47, opened 2020-09-03
Add documentation for the values.yaml


RyanW8:
It would be good to have some documentation on the values on what they do etc. For example https://github.com/helm/charts/tree/master/stable/prometheus#configuration this is a really nice way of documenting values showing how to set each value (YAML path), it's function & the default values.

PR #57, opened 2020-10-05
Discoverable Alertmanager + fixes + documentation

ts-mini:
Just attempting to contribute some of our changes to this chart upstream:

~~1) We're running multi-tenant - so we wanted to have ingress routes per scopeID when auth is enabled.  I've added this under distributor - but I'm not sure if this is the right place (as it ultimately creates a new route pointing to the nginx service)~~
2) I've added more configuration to the nginx service to make some of the debug/routes easier to hit from ingress.
~~3) Added a per-scope-id alertmanager path into nginx, configurable to alertmanager.ui.scopeids~~
4) added documentation
5) adjusted alertmanager discovery
I'm very flexible on implementation details - just interested in getting these style feature set upstream - so let me know if there are stylistic tweaks needed

[And added some readme things](https://github.com/cortexproject/cortex-helm-chart/issues/47)


templates/nginx-config.yaml
```diff
@@ -20,41 +20,82 @@ data:
...
+        location = /alerts/{{ $rootScope.scopeName }} {
+          proxy_set_header X-Scope-OrgID {{ $rootScope.scopeID }};
+          proxy_pass      http://{{ template "cortex.fullname" $root }}-alertmanager.{{ $root.Release.Namespace }}.svc.cluster.local:{{ $root.Values.config.server.http_listen_port }}/api/prom/alertmanager/;
+        }
+
+        location ~ /alerts/{{ $rootScope.scopeName }}/(.*) {
+          proxy_set_header X-Scope-OrgID {{ $rootScope.scopeID }};
+          proxy_pass      http://{{ template "cortex.fullname" $root }}-alertmanager.{{ $root.Release.Namespace }}.svc.cluster.local:{{ $root.Values.config.server.http_listen_port }}/api/prom/alertmanager/$1;
+        }
+{{- end }}
```
rsteneteg, line 89: Since we are setting the orgID header based on the map, could we add a similar location rule that we have for the other endpoints to the alertmanager ui?

templates/ingress-distributor.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: {{ template "cortex.fullname" $root }}-{{ .scopeName }}
+  labels:
+    app: {{ template "cortex.name" $root }}
+    chart: {{ template "cortex.chart" $root }}
+    release: {{ $root.Release.Name }}
+    heritage: {{ $root.Release.Service }}
+  {{- with .annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+    nginx.ingress.kubernetes.io/configuration-snippet: |
```
rsteneteg, line 23: see my other comment about the map in the nginx config, based on that I don't think we need ingress resources for the distributors.
I would also try to avoid using nginx specific ingress annotations since users may use other ingress controllers.

rsteneteg (commented):
Thanks for your contributions, had a few questions.

Really good start with the chart values in the readme

ts-mini:
@rsteneteg I will work on your comments asap! thanks for the feedback

templates/ingress-distributor.yaml
khaines: I agree with both of Roger's items here. We need to keep ingress resource templates agnostic to the kind of controller used, plus in this case, we can use the Nginx config settings which negates the need for another ingress resource.

templates/nginx-config.yaml
```diff
@@ -20,41 +20,82 @@ data:
...
       sendfile     on;
       tcp_nopush   on;
       resolver {{ default "kube-dns.kube-system.svc.cluster.local" .Values.nginx.config.dnsResolver }};
-
+{{- if .Values.config.auth_enabled }}
+      map $http_x_org_scopeid $scopeid {
+            default   $http_x_org_scopeid;
+            ""        "{{ .Values.auth.defaultScope }}";
+      }
+{{- end }}
```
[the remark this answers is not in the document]
khaines: Something to be aware of: authenticating/mapping requests to OrgID's is a concern cortex has aimed to keep separate from any of the components. It is far too easy to leak in implementations specific details into this area, so the choice in the past is to keep them external.  

I really appreciate the thought that went into using a config driven way with the built-in nginx deployment to support a form of multi-tenancy. I just have to push back on it because it is a very implementation-specific approach to multitenancy.

README.md
```diff
@@ -49,3 +49,573 @@ To upgrade Cortex use the following command:
...
+
+## Chart Requirements
+
+| Repository | Name | Version |
+|------------|------|---------|
+| https://kubernetes-charts.storage.googleapis.com/ | memcached | 3.2.3 |
+| https://kubernetes-charts.storage.googleapis.com/ | memcached | 3.2.3 |
+| https://kubernetes-charts.storage.googleapis.com/ | memcached | 3.2.3 |
+
+## Chart Values
```
khaines, line 63: Thank you so much for taking the time to document these! It's been a long pending item.

khaines (changes requested):
Thank you @ts-mini for the work here. This PR has a number of changes in it, some of which I think are ready to merge now and others I don't think are a good fit. I'm going to try my best to classify them:

- Documentation: Huge thank you for creating the values documentation
- Alertmanger discovery: There's some feedback here, but this is a great addition
- Nginx tweaks/fixes: Updates for correcting ports, etc ... most welcome
- Multi-tenancy mapping: See my comment about this being a concern we've wanted to keep external from cortex. I really like your thoughts and approach to this, but I don't think it is a good fit for bringing into the chart.

templates/nginx-config.yaml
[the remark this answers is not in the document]
ts-mini: Understood - I look forward to [this from "upstream" ](https://cortexmetrics.io/docs/roadmap/#auth-gateway).  I think what I will do is make a multi-tenant "add-on" chart then, that installs the nginx + distributor ingresses programmatically ( I think I can do this pretty easily and cleanly) and strip it out of this PR.

ts-mini:
I've made some changes - stripping out the stuff per @khaines comments.  I think this PR is a bit easier to digest now too - seeing as that complexity is gone.

I will work on making a secondary chart for our use that implements the multitenant separately from this chart - this will be a good path forward I think.  I understand the implications of disclosing implementation details - but may make this 'additional' chart available for anyone that wants to follow in my footsteps: deploy cortex, then deploy MT setup 'our/my way' - with the standard caveats of course.  If there's room in cortexproject for specific implementations - I wouldn't mind having it live here if y'all are ok with it, but that can be a future conversation.

rsteneteg:
LGTM

khaines (approved):
Thanks for making the adjustments @ts-mini ! This is a great addition.

LGTM also

PR #63, opened 2020-10-27
Add cluster DNS domain value

drewbowering:
When the Kubernetes cluster isn't using the default "cluster.local" DNS domain, deployments fail. This PR allows the DNS domain to be configurable, leaving "cluster.local" as the default.


rsteneteg:
Thank you for fixing this, LGTM

khaines:
LGTM! Thank you for the contribution

PR #68, opened 2020-10-28
Service Monitors and Compactor Service

ts-mini:
Seeing as this is included by [default in the Loki chart](https://github.com/grafana/loki/blob/master/production/helm/loki/values.yaml#L203) - I figure this might not be too opinionated of an addition.

I updated the docs as well

edit:  I discovered a alertmanager clustering bug (in the replicas code) that I also fixed

### DIFF
--- a/templates/alertmanager-statefulset.yaml
+++ b/templates/alertmanager-statefulset.yaml
@@ -19,7 +19,7 @@ spec:
       app: {{ template "cortex.name" . }}-alertmanager
       release: {{ .Release.Name }}
   updateStrategy:
-    {{- toYaml .Values.alertmanager.statefulStrategy | nindent 4 }}          
+    {{- toYaml .Values.alertmanager.statefulStrategy | nindent 4 }}
   serviceName: {{ template "cortex.fullname" . }}-alertmanager
   {{- if .Values.alertmanager.persistentVolume.enabled }}
   volumeClaimTemplates:
@@ -85,6 +85,8 @@ spec:
         {{- toYaml .Values.alertmanager.tolerations | nindent 8 }}
       terminationGracePeriodSeconds: {{ .Values.alertmanager.terminationGracePeriodSeconds }}
       volumes:
+        - emptyDir: {}
+          name: tmp
         - name: config
           secret:
             {{- if .Values.useExternalConfig }}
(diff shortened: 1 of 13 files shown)


khaines (commented):
Adding the option for the Prom Operator resources is handy. Can you please rebase your branch against the latest?

templates/alertmanager-statefulset.yaml
```diff
@@ -85,6 +85,8 @@ spec:
         {{- toYaml .Values.alertmanager.tolerations | nindent 8 }}
       terminationGracePeriodSeconds: {{ .Values.alertmanager.terminationGracePeriodSeconds }}
       volumes:
+        - emptyDir: {}
```
khaines, line 88: I would recommend avoiding specifying non-cortex path mounts here. I get the intention, but it prevents others from customizing /tmp in this setup. There are overridable settings for volumes and mounts, which allow for this.

khaines: @ts-mini - Sorry, should have flagged this earlier, but I think this last item needs correction prior to merge

ts-mini: sounds good I'll get to this asap

ts-mini: if I remember correctly, it's cause the code inside of cortex writes to tmp no matter what, and it's not customizable, so if you use the default psp the file system is ro, making it fail when it validates configs. I think I should be able to make it customizable, and adjust to tmp in our setup

khaines: Yes, that might be fair that cortex uses /tmp in some cases (I'll have to go look where again), but in the PR you've fixed that solution to be an emptyDir. If others are already mapping /tmp to other volume types, this will potentially break their setup.

ts-mini: Honestly I look back at this change and idk what I was thinking. 1) sorry for conflating prs, 2) I totally glossed over the 'extraVolume' and 'extraVolumeMount' options.  I am just going to use those going forward - and while I think we could _solve_ this ro-filesystem issue within the chart - the real "problem" imho is this [section](https://github.com/cortexproject/cortex/blob/6be8c0772d9e497cd4c7df928932c5aa6546bb2f/pkg/alertmanager/api.go#L154).  I might just _attempt_ to make this change on the cortex side so one could specify the 'tmpDIr' path as a config param.

khaines: Thank you for being open to the feedback and adjusting the changes :) LGTM now for merging

khaines (approved):
LGTM

PR #75, opened 2020-11-12
Change the ingester service to headless

jcmoraisjr:
Memberlist discovery can be configured using DNS. This commit changes the ingester service to headless, so memberlist can properly build the ring.

### DIFF
--- a/templates/ingester-svc.yaml
+++ b/templates/ingester-svc.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "cortex.fullname" . }}-ingester
+  name: {{ template "cortex.fullname" . }}-ingester-headless
   namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "cortex.name" . }}-ingester
@@ -15,6 +15,7 @@ metadata:
     {{- toYaml .Values.ingester.annotations | nindent 4 }}
 spec:
   type: ClusterIP
+  clusterIP: None
   ports:
     - port: {{ .Values.config.server.http_listen_port }}
       protocol: TCP
--- a/values.yaml
+++ b/values.yaml
@@ -135,6 +135,11 @@ config:
     shard_by_all_labels: true
     pool:
       health_check_ingesters: true
+  memberlist:
+    join_members: []
+    ## add here the service name of the ingesters
+    ## if using memberlist discovery, eg:
+    #  - cortex-ingester-headless
   querier:
     active_query_tracker_dir: /data/cortex/querier
     query_ingesters_within: 12h


stokkie90:
Don't we need to change the config file cortex uses based on the target? or can we put all headless records in the `join_members` option. 

```yaml
    memberlist:
        bind_port: 7946
        abort_if_cluster_join_fails: false
        join_members: 
            - cortex-distributor-headless.cortex.svc.cluster.local:7946
            - cortex-ingester-headless.cortex.svc.cluster.local:7946
            - ...
        max_join_backoff: 1m
        max_join_retries: 10
        min_join_backoff: 1s
```

if you have an example config for me that works would be pretty nice. or we could put some examples in this repo.

jcmoraisjr:
The memberlist configuration is currently built in the `values.yaml` file so I don't think we can build it conditionally. Thinking a bit more about it, I'd change ingester service to always be headless and add it to the join members of a fixed memberlist config. If one choose memberlist, the configuration is there, if not the configuration wouldn't be used. Anyway I'm pretty new to Cortex and I'd follow your suggestions.

stokkie90:
We might also take a look at how the cortex-jsonnet is being build. i digged a little bit into it. but it seems like they are setting a label on every service and use that as a selector:

```jsonnet
    ringConfig+: {
      'ring.store': 'memberlist',
      'memberlist.abort-if-join-fails': false,
      'memberlist.bind-port': gossipRingPort,
      'memberlist.join': 'gossip-ring.%s.svc.cluster.local:%d' % [$._config.namespace, gossipRingPort],
    },
  
  ....

  local gossip_member_label = 'gossip_ring_member',

  distributor_deployment_labels+:: { [gossip_member_label]: 'true' },
  ingester_deployment_labels+:: { [gossip_member_label]: 'true' },
  querier_deployment_labels+:: { [gossip_member_label]: 'true' },

  // Don't use gossip ring member label in service definition.
  distributor_service_ignored_labels+:: [gossip_member_label],
  ingester_service_ignored_labels+:: [gossip_member_label],
  querier_service_ignored_labels+:: [gossip_member_label],

  // Headless service (= no assigned IP, DNS returns all targets instead) pointing to some
  // users of gossiped-ring. We use ingesters as seed nodes for joining gossip cluster.
  // During migration to gossip, it may be useful to use distributors instead, since they are restarted faster.
  gossip_ring_service:
    local service = $.core.v1.service;

    // backwards compatibility with ksonnet
    local servicePort =
      if std.objectHasAll($.core.v1, 'servicePort')
      then $.core.v1.servicePort
      else service.mixin.spec.portsType;

    local ports = [
      servicePort.newNamed('gossip-ring', gossipRingPort, gossipRingPort) +
      servicePort.withProtocol('TCP'),
    ];
    service.new(
      'gossip-ring',  // name
      { [gossip_member_label]: 'true' },  // point to all gossip members
      ports,
    ) + service.mixin.spec.withClusterIp('None'),  // headless service
}
```

Source: https://github.com/grafana/cortex-jsonnet/blob/master/cortex/gossip.libsonnet

khaines:
As a simple default config, you can specify the headless service as the memberlist.join_members value. If the services aren't set to use memberlist for kv store, the value has no meaning but is harmless.

Also, since this config is set in the values.yaml file, it does not support the logic features that the templates do.

jcmoraisjr:
Hi, thanks for the update. Do you mean add this to the values.yaml?

```yaml
config:
  ...
  memberlist:
    join_members:
      - cortex-ingester-headless.cortex.svc.cluster.local
```

In this case this would need svc ingester to be configured always as headless, so remove the `ingester.service.headless` option, otherwise the service wouldn't be found if `headless==false`.

Any other default value? My cluster didn't get ready without `cortex-store-gateway-headless.cortex.svc.cluster.local` in this list - which only makes sense if using block storage.

khaines:
`My cluster didn't get ready without cortex-store-gateway-headless.cortex.svc.cluster.local in this list - which only makes sense if using block storage.` Can you add more details to this? I'm running my staging cluster with block storage and I only have the ingester headless service as the memberlist "join_members" value.

As regards to always having the ingesters as a headless service vs a switch, I've been trying to find a reason for them to not have a headless service (it may have been added out of habit).

jcmoraisjr:
> Can you add more details to this?

It seems I cannot reproduce anymore, probably an issue with any other config which was fixed elsewhere.

Regarding the default join_member config: there is a convention to use `cortex` as the namespace but this could be overwritten - one just need to not configure the namespace and the kubeconfig's default would be used. I'd propose to use instead like the snippet above, if this seems good enough and you cannot see any issue with dns resolve configs out there.

```yaml
  memberlist:
    join_members:
      - cortex-ingester-headless
```

It works on my cluster, running flatcar and k8s 1.18.8.

> I've been trying to find a reason for them to not have a headless service

I'd follow this as well, but I'm pretty new to Cortex and I don't see the whole picture yet.

khaines:
To add some more twists to the naming, even `cortex-ingester-headless` isn't necessarily going to work in many cases, because the name is prefixed with the "release name". Perhaps it is best in this case we leave the default as "empty" with a commented section in the values.yaml that this needs to be filled in if using memberlist with an example or two within the comment block.

jcmoraisjr:
Just updated the PR to only move ingester service to headless as proposed. `values.yaml` has also a commented configuration snippet to add the ingesters as members of the ring if using memberlist.

khaines (approved):

PR #76, opened 2020-11-12
Fix settings labels and annotations for all services

kwangil-ha:
Add settings service type for nginx
Add headless service for compactor
Set new location for the stable repository
Fix syntax error

Signed-off-by: 하광일/TC_Core Engineering <kwangil_ha@yanolja.com>


templates/compactor-svc.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if eq .Values.config.storage.engine "blocks" -}}
```
khaines, line 1: What does a headless service provide for the compactor?

khaines (commented):
Once the merge conflicts are addressed , I think this is read for merge

PR #78, opened 2020-12-01
fix: nginx config using non-default clusterDomain

drewbowering:
- One occurrence of hard-coded "cluster.local" changed to clusterDomain

Signed-off-by: Drew Bowering <Drew.Bowering@epl.ca>


khaines (approved):
Great catch! Thank you

PR #85, opened 2020-12-03
Restructuring

langecode:
For what it's worth, thought I would provide you with this PR even if I acknowledge that it is probably way too big to be accepted - but here goes.

I was looking to do a production installation of Cortex and since we are normally using Helm charts I came across this charts from the Cortex project. I have introduced the following changes:

- Created a directory structure to reflect the different components of Cortex to have a more clean grouping of the manifests
- Using Helm templates for all labels (labels and the match part of the label sets)
- Changed labels to adhere to the [Kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) (and [Helm best practices](feature/PD-177-chart-clean-up))
- Removed the default Prometheus scrape labels to leave this to the user
- Fixed the pod disruption budget configuration so that each component can have its own pdb
- Added memcached configuration suitable when running blocks storage
- Added memcached configuration for frontend component
- Added Consul subchart
- Added example of values for running with blocks storage (`values-blocks.yaml`)
- Made it possible to disable the optional `configs` component - if running Alertmanager and Ruler with locale configuration, e.g., from ConfigMap

You can take it or leave it. If there are minor changes I am sure I can find a little time to fix some. Otherwise we just maintain our own fork - just thought I would give you the chance benefit from the time we spent on this.


khaines:
@langecode - I appreciate the time that went into this PR and there are some great contributions in it. However, this change is far too large and disruptive to be taken as a whole.

Almost each of your bullet points could be a PR on its own as they have their own merits and potential discussion.

If this isn't something you have time for, I understand. You've got some changes in here I've wanted to see for a bit, so I can pick those out and create PRs for them.

langecode:
I perfectly understand that. We just saw a number of issues that we felt needed to be corrected before we went into production with Cortex and instead of just keeping the changes to ourselves we decided to feed those back. Having to wait for all these changes to be merged into the upstream project just would take too long. I pretty sure I would not have the time to split all these changes up into individual PRs - but I do understand why you would like to do that. We will just to keep up with incoming upstream changes and hope your chart converge with the Helm and Kubernetes best practices over time. Thank you for a very nice product with Cortex - really seems to work well for us so far.

Issue #86, opened 2020-12-10
Alertmanager cluster peering fails due to wrong url 


kevinbheda:
Alertmanager logs says 
"Failed to resolve cortex-alertmanager-alertmanager-1.cortex-alertmanager-alertmanager.alertmanager.svc.cluster.local:9094: lookup  no such host"
Root cause,
The servicename used in alertmanager-statefulset.yaml is of cluster endpoint service. 
To fix the issue it should use headless service which will resolve the dns of individual pods by hostname.

Will raise a PR for the same.

PR #87, opened 2020-12-10
fix: alertmanager cluster peers not reachable due to wrong dns name

kevinbheda:
This fixes issues [86](https://github.com/cortexproject/cortex-helm-chart/issues/86)


khaines (approved):
LGTM, thank you @kevinbheda !

PR #91, opened 2020-12-17
deploy alertmanager headless service when StatefulSet is used

khaines:
currently, the headless service for the alertmanager is deployed in all configurations. However, it is only needed when the alert manager is a stateful set.


rsteneteg:
LGTM, thanks Ken

PR #92, opened 2020-12-17
Unified memcache settings into template function + add blocks cache

khaines:
Based on changes picked from #85, this unifies the cache settings into a template function and adds in support for Memcached instances when using blocks storage.

Signed-off-by: Ken Haines <khaines@microsoft.com>


khaines:
This has been running in staging over the holidays without issue and the cache resources are showing metrics of being utilized by cortex. Since there have been no objections to the PR, I'm going to merge this into master.

Issue #94, opened 2020-12-21
networking.k8s.io/v1beta1 Ingress is deprecated 


dilipkk-foyernet:
```sh
helm install cortex --namespace cortex cortex-helm/cortex
W1221 16:56:00.761473 2564402 warnings.go:67] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
Error: create: failed to create: namespaces "cortex" not found
```

jcmoraisjr:
You should also add `--create-namespace` if want helm do it for you.

PR #95, opened 2020-12-21
Update ingress.yaml 

dilipkk-foyernet:
Update apiVersion: networking.k8s.io/v1beta1 -> networking.k8s.io/v1 for Kubernetes versions 1.19+

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [ENHANCEMENT] Compactor service is no longer created if compactor is disabled. #82
 * [ENHANCEMENT] Headless service for alert manager is only enabled when the alert manager is deployed as a stateful set. #91
+* [ENHANCEMENT] Update apiVersion: networking.k8s.io/v1beta1 -> networking.k8s.io/v1 for Kubernetes versions 1.19+  #94
 
 ## 0.2.0 / 2020-10-25
 
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -1,7 +1,9 @@
 {{- if and .Values.ingress.enabled .Values.nginx.enabled -}}
 {{- $fullName := include "cortex.fullname" . -}}
 {{- $svcPort := .Values.nginx.http_listen_port -}}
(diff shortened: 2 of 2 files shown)


dilipkk-foyernet:
This fixes: https://github.com/cortexproject/cortex-helm-chart/issues/94

khaines:
Update the changelog as well, please. Otherwise this LGTM! thank you for the PR.

dilipkk-foyernet:
> Update the changelog as well, please. Otherwise this LGTM! thank you for the PR.

Done @khaines

langecode:
Saw this was merged - nice - however, just a few comments.

For the `semverCompare` function to work the hyphen should be a dot line 4 and 6 in `ingess.yaml` (see also [Semantic Version Functions](https://helm.sh/docs/chart_template_guide/function_list/#semantic-version-functions)).

The structure of the ingress manifest has changed with Kubernetes 1.19 so it needs to rendered according to whether its 1.19 or previous ingress (see [ingress.yaml](https://github.com/langecode/cortex-helm-chart/blob/master/templates/ingress.yaml#L40))

dilipkk-foyernet:
Thanks, @langecode, I just followed the pattern used previously and your point is valid. For now, we can keep it like this as it is working as expected with - also: https://semvercompare.azurewebsites.net/?version=1.14-0&version=1.14.1.

We need to refactor `.Capabilities.KubeVersion.GitVersion` to `.Capabilities.KubeVersion.Version` because of the following issue:
https://github.com/helm/charts/issues/20918, we can correct this error also at the same time

dilipkk-foyernet:
I am not too sure about Go version, need to check that, yes in your fork it is showing as `>=1.19.0`.

langecode:
Indeed, and I think I did that because we are running on a 1.19.x cluster and I got the deprecation warning mentioned in #94 testing out the deployment. Getting the deprecation warning Kubernetes is actually rewriting the resource `kind` which then causes Helm to loose the relation between what is deployed and what is actually there - and this can cause a pretty messy clean up.. been there, done that :-)

dilipkk-foyernet:
https://play.golang.org/p/deMXSnOTHyZ
Tested with golang implementation also, it is working as expected.

PR #101, opened 2021-01-08
as nginx uses http 1.0 by default, which is very old and istio servic…

Nitesh-vaidyanath:
…e mesh wont support 1.0 without making changes to istio config. Instead of chaning istio config we can set 1.1 by default in nginx config file


eMurgaNd:
I also need this change.

PR #103, opened 2021-01-20
Expose new HTTP API paths for remote-write and Query frontend to nginx

AkarshSatija:
- Adds new remote write API to nginx. 
    - `/api/v1/push`
    - API ref: https://cortexmetrics.io/docs/api/#remote-write
- Adds new query frontend APIs to nginx
    - `/prometheus/api/v1/label/.*`
    - `^/prometheus/api/v1/(read|metadata|labels|series|query_range|query)`
    - API ref: https://cortexmetrics.io/docs/api/#querier--query-frontend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Expose new HTTP API paths for remote-write and Query frontend to nginx #103
 * [ENHANCEMENT] Compactor service is no longer created if compactor is disabled. #82
 * [ENHANCEMENT] Headless service for alert manager is only enabled when the alert manager is deployed as a stateful set. #91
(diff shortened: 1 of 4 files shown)


bhiravabhatla:
@AkarshSatija  - prometheus_http_prefix is configurable and is defaulted to prometheus. 
/prometheus/api/v1/label/.*  is actually /<prometheus_http_prefix>/api/v1/label/.* ( where prometheus_http_prefix is defaulted to prometheus) - we should ideally pick this from .Values.config.api.prometheus_http_prefix if its present. if not use **prometheus**.

https://cortexmetrics.io/docs/configuration/configuration-file/

AkarshSatija:
Thanks @bhiravabhatla. This totally makes sense. I have made the changes accordingly

Issue #116, opened 2021-02-04
Helm Chart tags not working


cabrinha:
Setting the tag to true does not create the memcached-blocks charts:

```yaml
tags:
  blocks-storage-memcached: true
```

Can we just use the `condition` bool instead?

```
❯ helm version
version.BuildInfo{Version:"v3.5.1", GitCommit:"32c22239423b3b4ba6706d450bd044baffdcf9e6", GitTreeState:"dirty", GoVersion:"go1.15.7"}
```

PR #117, opened 2021-02-05
Fix chart conditional dependencies

cabrinha:
Just use a conditional for the memcached-blocks charts

solves https://github.com/cortexproject/cortex-helm-chart/issues/116


cabrinha:
I was using this chart as a sub-chart and need to set tags in the cortex subchart as well as the root of my parent chart.

PR #126, opened 2021-02-24
Fix serviceName of store gateway StatefulSet

xeniumlee:


khaines:
@xeniumlee - Thank you for the PR! The change itself looks good, but can you please rebase your branch against master and then signoff on the amended change?

PR #128, opened 2021-02-25
nginx: allow basic auth

Eraac:


khaines:
I think I understand your intention with this change, and I appreciate the time you put in creating this PR. However, I'm not certain this is a change we want to bring into the main codebase. Tenant/User Authentication/Authorization is strictly not a part of the core project and while there's been an NGINX component in samples & this helm chart, it was not intended for long term (ie: external/production) usage. Adding in Basic auth would blur this line, so for now I'm going to close the PR.

Issue #119, opened 2021-02-09
Support AWS LB Controller Ingress


SudoBrendan:
Greetings,

The current chart does not support all ingress controller schemas, including [AWS Load Balancer Controller](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html). The main limitation currently is InstanceIP mode, where [Ingress objects must be backed by a NodePort type Service](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/how-it-works/#instance-mode). It'd be a good idea to also support additional paths in the Ingress object so [SSL redirection can happen](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.1/guide/tasks/ssl_redirect/#example-ingress-manifest).

Might I suggest adding the following values:
1. `nginx.service.type`: sets the type on the Service fronting nginx
2. `ingress.extraPaths`: raw YAML for additional paths to be added to the Ingress - [here's how grafana did this](https://github.com/grafana/helm-charts/blob/1822ba00099b96b89201517369a49360f5f8b921/charts/grafana/templates/ingress.yaml#L40-L42)

PR #131, opened 2021-03-01
nginx: service type configurable

Eraac:


khaines:
I'm not clear on the intention with this change. Could you elaborate some more on the functionality gained? Aside from ClusterIP, I'm not sure the other types fit well: Headless could work, but I'm not seeing the benefit. The remaining service types are all for accepting external traffic and an Ingress Resource (which exists in the chart) is a better match for that scenario.

Eraac:
When you use a GCP LB (or AWS I think https://github.com/cortexproject/cortex-helm-chart/issues/119) you already have an Ingress object, which need a service with a NodePort

AkarshSatija (approved):

rohilsurana:
Bump.
We have exact same use case where we would like to use the service with a GCP Internal LB.

PR #134, opened 2021-03-02
query-frontend: set memcache config

Eraac:


khaines (approved):
LGTM! thank you for adding this.

PR #139, opened 2021-03-22
Fix dependencies in charts dir to reflect recent update in 0.4.0

### LINKED ISSUE #135: Release 0.4.0 breaks memcached functionality
Installing / upgrading cortex to chart `v0.4.0` seems to break the memcached functionality (memcached gets uninstalled which causes cortex to fail if I upgrade to `v0.4.0`) . Somehow the memcached dependencies are not installed even when declaring `tags.blocks-storage-memcached=true`. To verify this, I ran the following command (I know that more configuration options need to be present to properly configure cortex for block storage, but this command should be sufficient to reproduce the issue):

```bash
helm install --dependency-update --verify=false --dry-run cortex --namespace cortex --set tags.blocks-storage-memcached=true cortex-helm/cortex | grep -i memcached
```

When I pull the chart, update the dependencies and then install it from the local folder everything seems to work:

```bash
helm pull cortex-helm/cortex
tar xzf cortex-0.4.0.tgz
helm dep update cortex/
helm install --dependency-update --verify=false --dry-run cortex --namespace cortex --set tags.blocks-storage-memcached=true cortex/ | grep -i memcached
```

Looks like the dependencies in the `v0.4.0` release are missing somehow.

gburton1:
This fixes #135. I tested it on my fork and it deployed successfully with all of the expected caches.

Generation commands:

```
$ helm package . --version 0.4.1 --destination ./docs --dependency-update
Successfully packaged chart and saved it to: docs/cortex-0.4.1.tgz

$ helm repo index . --url https://cortexproject.github.io/cortex-helm-chart/
```


khaines:
Thanks @gburton1 for updating this chart! The effort is really appreciated here.

Skaronator:
Any plans to make a new release available?

PR #142, opened 2021-04-13
Upgrade memcached and add sa token mount value

ryanprobus:
* Upgrades memcached helm chart from 5.5.1 to 5.13.0 which adds support
  for specifying service account association and creation
* Adds service account value for whether to automount the token

Signed-off-by: Ryan Probus <ryan.probus@appian.com>

### DIFF
--- a/Chart.lock
+++ b/Chart.lock
@@ -1,24 +1,24 @@
 dependencies:
 - name: memcached
   repository: https://charts.bitnami.com/bitnami
-  version: 5.5.1
+  version: 5.13.0
 - name: memcached
   repository: https://charts.bitnami.com/bitnami
-  version: 5.5.1
+  version: 5.13.0
 - name: memcached
   repository: https://charts.bitnami.com/bitnami
-  version: 5.5.1
+  version: 5.13.0
 - name: memcached
   repository: https://charts.bitnami.com/bitnami
-  version: 5.5.1
+  version: 5.13.0
 - name: memcached
   repository: https://charts.bitnami.com/bitnami
-  version: 5.5.1
+  version: 5.13.0
 - name: memcached
   repository: https://charts.bitnami.com/bitnami
-  version: 5.5.1
+  version: 5.13.0
 - name: memcached
   repository: https://charts.bitnami.com/bitnami
-  version: 5.5.1
-digest: sha256:944a782569cae3c73869587c9efbd85dd4841fb28d648fdaf6fd72e9c0dfd7d2
-generated: "2021-02-19T16:22:13.325783-08:00"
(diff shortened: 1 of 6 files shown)


Chart.lock
```diff
@@ -1,24 +1,24 @@
 dependencies:
 - name: memcached
   repository: https://charts.bitnami.com/bitnami
-  version: 5.5.1
+  version: 5.10.1
```
ShuzZzle, line 4: Sorry for the wait time. Do you want to bump that up to 5.12.2? (the newest version)

However bitnami/memcached seems to have some quite severe [CVE's](https://artifacthub.io/packages/helm/bitnami/memcached?modal=security-report) ( i am guessing the old version does too...)

Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 0.4.1
+version: 0.5.0
```
ShuzZzle, line 2: Please don't bump the Chart version on your own

Chart.yaml
```diff
@@ -15,39 +15,39 @@ sources:
 dependencies:
   - name: memcached
     alias: memcached
-    version: 5.5.1
+    version: 5.10.1
```
ShuzZzle, line 18: See first comment....duh :D

ShuzZzle (changes requested):

ShuzZzle:
@ryanprobus  Also please document your Change in the CHANGELOG.md

>  CHANGELOG.md updated - the order of entries should be [CHANGE], [FEATURE], [ENHANCEMENT], [BUGFIX]

cabrinha:
If there is a better memcached chart we can go with, let's take a look at it.

ShuzZzle:
> If there is a better memcached chart we can go with, let's take a look at it.

I wish...I think we just have to suck it up until maybe something like redis is supported or whatever

ryanprobus:
@ShuzZzle Sorry, was on vacation past couple weeks. Removed chart version bump, added to changelog and changed memcached upgrade from 5.5.1 to 5.13.0 (currently the newest published version).

Chart.lock
ryanprobus: Upgraded to 5.13.0 (currently highest published version) since it had been a couple weeks since this comment.

Chart.yaml
ryanprobus: Removed.

Chart.yaml
ryanprobus: See first comment's response =)

ShuzZzle (approved):

PR #144, opened 2021-04-28
Refactor chart & change labels to best practices

Skaronator:
This PR includes the following changes:
* modify labels to the recommended [Helm best practices](https://helm.sh/docs/chart_best_practices/labels/)
* Allow disabling of cortex alertmanager, configs & ruler
* Configure alerts/rules via configmap
* Create dedicated memberlist service for all microservices 

## Labels

old labels
```yaml
app: {{ template "cortex.name" . }}-alertmanager
chart: {{ template "cortex.chart" . }}
release: {{ .Release.Name }}
heritage: {{ .Release.Service }}
```

new lables:

```yaml
helm.sh/chart: cortex-0.5.0
app.kubernetes.io/name: cortex
app.kubernetes.io/instance: cortex
app.kubernetes.io/version: "v1.8.1"
app.kubernetes.io/managed-by: Helm
app.kubernetes.io/component: alertmanager
```

## Disable alertmanager / configs / ruler

You can now disable the alertmanager, configs & ruler via values. They are still enabled by default.

```yaml
alertmanager:
  enabled: false

ruler:
  enabled: false

configs:
  enabled: false

```

## Configure rules via configmap

You can now configure your rules via configmap if you don't want to use the configs or a s3 bucket:

```yaml
ruler:
  directories:
    tenant_foo:
      rules1.txt: |
        groups:
          - name: should_fire
            rules:
              - alert: HighPercentageError
                expr: |
                  sum(rate({app="foo", env="production"} |= "error" [5m])) by (job)
                    /
                  sum(rate({app="foo", env="production"}[5m])) by (job)
                    > 0.05
                for: 10m
                labels:
                  severity: warning
                annotations:
                  summary: High error rate
```

## Dedicated memberlist service

Part of the new labels it make sense to actual use a dedicated service for the memberlist.

```yaml
config:
  memberlist:
    join_members: 
    - cortex-memberlist
```
This service only includes microservices that actually support memberlist. This is done with the following label:
`app.kubernetes.io/part-of: memberlist`

----
Due to all these changes this PR creates breaking changes but since we are still at 0. major version it should be fine ;)

Most of these changes are inspired from the loki chart which I really like: https://github.com/grafana/helm-charts/tree/main/charts/loki-distributed/templates


amouhadi:
I came across your PR and realized it's similar to what I've been working on. Our approaches are a little different. I added a sidecar container to the Ruler and AlertManager pod that watches the k8s API and mounts ConfigMaps dynamically, while you are creating the ConfigMaps in the chart itself from chart values. Have you considered an approach like mine that doesn't couple the rule changes to chart deployments? The Grafana chart does it with the same sidecar approach, which is where I got the idea.
https://github.com/cortexproject/cortex-helm-chart/pull/150

templates/compator/_helpers-compator.tpl
```diff
@@ -0,0 +1,23 @@
+
+{{/*
+compator fullname
```
babilen5, line 3: Should this be `compactor` everywhere?

babilen5 (changes requested):

templates/compator/_helpers-compator.tpl
Skaronator: Yep you're correct. Fixed in latest commit.

ShuzZzle:
thanks for the PR! I love the changes, I merged some small fixxes first so we can have some head room for the big one (yours)

templates/alertmanager/_helpers-alertmanager.tpl
```diff
@@ -0,0 +1,23 @@
...
+{{- define "cortex.alertmanagerFullname" -}}
+{{ include "cortex.fullname" . }}-alertmanager
+{{- end }}
+
+{{/*
+alertmanager common labels
+*/}}
+{{- define "cortex.alertmanagerLabels" -}}
+{{ include "cortex.labels" . }}
+app.kubernetes.io/component: alertmanager
```
ShuzZzle, line 14: Excellent Idea, especially since its all the same binary

templates/podsecuritypolicy.yaml
```diff
@@ -3,12 +3,8 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
```
ShuzZzle, line 3: PSP are deprecated, honestly we should just drop it

templates/role.yaml
```diff
@@ -3,12 +3,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
```
ShuzZzle, line 3: Looks like the only reason for the role is PSP which is deprecated anyway. So might be valid to get rid of it as well

templates/rolebinding.yaml
```diff
@@ -3,12 +3,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
```
ShuzZzle, line 3: see comment above

templates/svc-memberlist.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+    {{- include "cortex.labels" . | nindent 4 }}
+spec:
+  type: ClusterIP
+  clusterIP: None
+  ports:
+    - port: 7946
+      protocol: TCP
+      name: http-metrics
+      targetPort: http-metrics
```
ShuzZzle, line 14: why are you sending port 7946 to port 8080?
Also: shouldn't the memberlist components be exposing port 7946?

templates/svc-memberlist.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: {{ include "cortex.fullname" . }}-memberlist
+  labels:
+    {{- include "cortex.labels" . | nindent 4 }}
+spec:
+  type: ClusterIP
+  clusterIP: None
+  ports:
+    - port: 7946
+      protocol: TCP
+      name: http-metrics
```
ShuzZzle, line 13: Should be renamed to grpc / gossip / or whatever

values.yaml
```diff
@@ -832,6 +892,7 @@ table_manager:
   env: []
 
 configs:
+  enabled: true
```
ShuzZzle, line 895: I think its heading more in a remote direction (s3), might be worth disabling it by default

ShuzZzle (changes requested):
Please rebase against the current master. Also if you want my feedback or go  through it again with me, then let me know. Nice Work!

You can find me on the official cortex slack (CNCF/#cortex) @ShuzZzle

templates/podsecuritypolicy.yaml
Skaronator: Removed ;)

Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
-version: 0.4.1
-appVersion: v1.7.0
+version: 0.5.0
```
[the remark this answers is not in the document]
Skaronator: Reverted

templates/svc-memberlist.yaml
Skaronator: You're right. Weird that it worked without... fixed :+1:

templates/svc-memberlist.yaml
Skaronator: You're right. Changed to gossip

values.yaml
Skaronator: :+1:  disabled by default

templates/alertmanager/alertmanager-dep.yaml
```diff
@@ -77,6 +74,10 @@ spec:
             - name: http-metrics
               containerPort: {{ .Values.config.server.http_listen_port }}
               protocol: TCP
+          {{- if semverCompare ">=1.20-0" .Capabilities.KubeVersion.GitVersion }}
+          startupProbe:
```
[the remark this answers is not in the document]
Skaronator: I'm not sure whats the best way to handle this case. I didn't want to do a `if {{ .Values.alertmanager.startupProbe.create }}` because the `create` value would break the `{{- toYaml .Values.alertmanager.startupProbe | nindent 12 }}` a line below.

I think we have the following options:

## Use semverCompare

This approach would enable `startupProbe` automatically without further work by the user. Only downside is that users with beta feature flag and kubernetes 1.18 / 1.19 cannot use it.

## Force startupProbe

We simply remove the if and optimize the livenessProbe defaults to only work together with startupProbe. This would likely break a lot of installs since a lot of user still aren't using 1.20 or 1.18 with beta features.


## Leave startupProbe defaults empty

We can leave startupProbe defaults empty in the values.yaml and do a `if {{ .Values.alertmanager.startupProbe }}` to check if the values are set or empty.

This means by default startupProbe is never used. The user has to supply values and adjust the livenessProbe to work properly.

## Leave startupProbe defaults NOT empty

This is similar to the option above with the same `if` block but we now provide default values.yaml for the startupProbe. 

User that use a Kubernetes Version that does not support `startupProbe` need to set `.Values.alertmanager.startupProbe` to `null` and likely need to increase the livenessProbe `failureThreshold`.

---------

What option would you prefer @ShuzZzle. I'd go with the first or the last option.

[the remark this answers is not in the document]
ShuzZzle: I believe startupProbe was introduced in 1.16. I don't think we should aim to support k8 Version older than the officialy supported versions which are currently 1.21-1.19

Therefore startupProbe entered Beta in 1.18 and from there on it was enabled by default! [Feature-Gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/)

So I think we can simply remove the check ( I am gonna ask a Maintainer to confirm this)

templates/svc-memberlist.yaml
ShuzZzle: I think memberlist only uses the service for discovery, the port might be hardcoded

ShuzZzle:
I see @Skaronator you also deleted "templates/distributor-svc-headless.yaml", I understand why but there is actually a use case where this could be helpful. You can use this to talk to the distributors directly via grpc and protobuf, might be of interest and doesn't actually hurt anyone

Skaronator:
I've moved the `templates/distributor-svc-headless.yaml` to `templates/distributor/distributor-svc-headless.yaml`. The diff looks just weird.

I need to fix the pod disruption budget and adjust the startupProbe/livenessProbe defaults before this can be merged. Give me 2-3 hours :)

ShuzZzle:
> I've moved the `templates/distributor-svc-headless.yaml` to `templates/distributor/distributor-svc-headless.yaml`. The diff looks just weird.
> 
> I need to fix the pod disruption budget and adjust the startupProbe/livenessProbe defaults before this can be merged. Give me 2-3 hours :)

Oh sorry, my bad. Yeah no worries, don't rush yourself. I want @cabrinha to have a look at it before I merge it anyway

templates/service-memberlist.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
ShuzZzle, line 1: maybe we should rename this file to
service-memberlist-headless.yaml or
svc-memberlist-headless.yaml

since that was naming scheme

Skaronator:
Alrighty I tested most of the features in our dev environment in aws. Everything seems to work so far.

templates/alertmanager/alertmanager-dep.yaml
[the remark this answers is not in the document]
Skaronator: I've removed the check and adjust the default values of the livenessProbe to be more in line with the startupProbe workflow. Also bumped the minimum version to 1.18 in the chart.yaml and removed pre-1.14 Ingress API.

ShuzZzle:
> Alrighty I tested most of the features in our dev environment in aws. Everything seems to work so far.

I just realized. We also need to document what actually changed in the CHANGELOG.md. Should be a couple of entries

Phew so much todo. But we will get there!

Skaronator:
I've added the changes to the changelog. Hopefully got everything covered.

values.yaml
```diff
@@ -20,9 +20,9 @@ tags:
   blocks-storage-memcached: false
 
 ingress:
-  enabled: true
+  enabled: false
```
ShuzZzle, line 23: Is there a reason why we disable that by default? I feel like it should be easy to get it going with some decent default settings. But I don't have a strong opinion about it.

Skaronator: I can't remember the last chart I had used where this was enabled by default. Personally I'd like to have it `false` by default because not everyone need a ingress and the placeholder values would cause issues if you don't configure it properly.

Looking at other charts they have it all disabled by default:

1. [Loki](https://github.com/grafana/helm-charts/blob/7e5cce564a51fd29f39019224e5cf1c9f32d4996/charts/loki-distributed/values.yaml#L571-L573)
2. [Grafana](https://github.com/grafana/helm-charts/blob/7e5cce564a51fd29f39019224e5cf1c9f32d4996/charts/grafana/values.yaml#L179-L180)
3. [Consul](https://github.com/bitnami/charts/blob/a51a77c505509cd0c2251e41803ec617208f4457/bitnami/consul/values.yaml#L392-L395)
4. Even [Wordpress](https://github.com/bitnami/charts/blob/a51a77c505509cd0c2251e41803ec617208f4457/bitnami/wordpress/values.yaml#L498-L501)

ShuzZzle (approved):
LGTM!

ShuzZzle:
@Skaronator 

> Error:  templates/nginx/nginx-ingress.yaml: the kind "extensions/v1beta1 Ingress" is deprecated in favor of "networking.k8s.io/v1beta1 Ingress"

I fixxed that by simply removing k8 1.18 support since it's not supported anymore anyway.
Also we now need to adjust ci/test-values.yml since some variables changed, right?

Skaronator:
Probably only this line if the CI does a actual deployment:

https://github.com/cortexproject/cortex-helm-chart/blob/5888a3c1007691056bbb068f62f75591dd4916d5/ci/test-values.yaml#L45

```diff
- cortex-ingester-headless.cortex-green
+ cortex-memberlist.cortex-green
```

ShuzZzle:
> Probably only this line if the CI does a actual deployment:
> 
> https://github.com/cortexproject/cortex-helm-chart/blob/5888a3c1007691056bbb068f62f75591dd4916d5/ci/test-values.yaml#L45
> 
> ```diff
> - cortex-ingester-headless.cortex-green
> + cortex-memberlist.cortex-green
> ```

Yep you should be able to use your new feature :eyes:

ShuzZzle:
@Skaronator  where does cortex-ingester-headless.**cortex-green** come from? Is that supposed to be kubernetes DNS Discovery but then the namespace should be cortex

Skaronator:
Absolutly no clue. Was wondering that myself but you can just remove it. We don't need the namespace in that (like in line 41)

ShuzZzle:
> Absolutly no clue. Was wondering that myself but you can just remove it. We don't need the namespace in that (like in line 41)

you should join the cncf slack and the #cortex channel there. then we can talk there. I am gonna debug it tomorrow in a fresh cluster

ShuzZzle:
no idea how the test-values ever worked, but it's definitely correct now

PR #148, opened 2021-05-05
Add headless service for distributor

lahsivjar:
Create headless service for distributor to allow GRPC load balancing when using HTTP over GRPC for communicating with distributor.


ShuzZzle (commented):
LGTM! However It might be a good idea to have an option to disable the distributor headless service so the user doesn't get confused

Edit: I just noticed thats not what we did apparently with the query_frontend services, so might not be that important

ShuzZzle (approved):

cabrinha (approved):

PR #149, opened 2021-05-06
Add tenant overrides config

puneeth8994:
Creates a secret that stores the override configs data.
Enables setting specific limits to different tenants.

Ref: https://cortexmetrics.io/docs/guides/overrides-exporter/


ShuzZzle:
Thanks,

I might be of interest to some people that we also automatically mount the override config and provide the necessary arguments when enabling it via tenantOverridesConfig.enabled

PR #157, opened 2021-06-04
fix pdb manifests

### LINKED ISSUE #151: podDisruptionBudget option does not work
`podDisruptionBudget` option is declared inside each service:
```
ingester:
  ...
  podDisruptionBudget: {}
  ...
```

but the template expects a global value:
```
{{- if .Values.podDisruptionBudget -}}
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: {{ template "cortex.fullname" . }}
...
```

AndreZiviani:
**What this PR does**:
Fixes the PodDisruptionBudget logic.

The manifest was probably created based on `target: all` mode, this PR creates a template for each component.

Fixes #151 

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [BUGFIX] Fixed `priorityClassName` in alertmanager deployment configuration. #155
 * [BUGFIX] Replacing deprecated (in Cortex 1.9) configuration with appropriate replacements (max_look_back_period -> max_query_lookback & compress_responses -> response_compression_enabled). #154
 * [BUGFIX] Added missing ServiceMonitors `additionalLabels` placeholders in `values.yaml`. #146
+* [BUGFIX] Fix PodDisruptionBudget templates. #156
 
 ## 0.4.1 / 2021-03-22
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * [ENHANCEMENT] Expose `client_max_body_size` config for nginx max request body size #137
 * [ENHANCEMENT] Adding option to add custom headers (ex. X-Scope-OrgID) to NGINX from values.yaml (key `nginx.config.setHeaders`). #127
+* [BUGFIX] Fix PodDisruptionBudget templates. #156
```
ShuzZzle, line 7: I dont really know if that falls into the category BUGFIX but the PR number is definitely wrong. Should be #157

ShuzZzle (changes requested):

CHANGELOG.md
AndreZiviani: Yeah, I wasn't sure what to use, maybe enchancement?

> the PR number is definitely wrong. Should be #157
:facepalm:

ShuzZzle:
Thank you so much!
However #144 (don't really know why) just did the same thing you did. And we are currently in the process of merging this PR. So even though everything is correct here, I might have to close it in favor of #144

:disappointed:

AndreZiviani:
Looks like that implementation has some improvements over mine, I don't mind using that instead. Thanks for the review!

PR #158, opened 2021-06-06
Enable / Disable Security Context

### LINKED ISSUE #140: Enable / Disable Security Context
I think it will be better if we can remove the security context from pods when we don't want them. if you agree I can do this with the following code snippet just like the Bitnami charts.

```
{{- if .Values.podSecurityContext.enabled }}
securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
{{- end }}
```

1995parham:
This PR makes the security context removable from Kubernetes YAML files, it means you can really disable them in case of not wanting them. closes #140


ShuzZzle:
@1995parham Somebody didn't use the PR template :D

Please fill out the changelog:

templates/alertmanager-dep.yaml
```diff
@@ -43,8 +43,10 @@ spec:
     {{- if .Values.alertmanager.priorityClassName }}
       priorityClassName: {{ .Values.alertmanager.priorityClassName }}
     {{- end }}
-      securityContext:
-        {{- toYaml .Values.alertmanager.securityContext | nindent 8 }}
+
```
ShuzZzle, line 46: can you remove the blank line, please?

ShuzZzle (changes requested):

1995parham:
Thanks for your comments. I have fixed them.

CHANGELOG.md
```diff
@@ -2,48 +2,48 @@
...
-* [BUGFIX] Fixed `priorityClassName` in alertmanager deployment configuration. #155
-* [BUGFIX] Replacing deprecated (in Cortex 1.9) configuration with appropriate replacements (max_look_back_period -> max_query_lookback & compress_responses -> response_compression_enabled). #154
-* [BUGFIX] Added missing ServiceMonitors `additionalLabels` placeholders in `values.yaml`. #146
+- [ENHANCEMENT] Expose `client_max_body_size` config for nginx max request body size #137
+- [ENHANCEMENT] Adding option to add custom headers (ex. X-Scope-OrgID) to NGINX from values.yaml (key `nginx.config.setHeaders`). #127
+- [ENHANCEMENT] Headless service for distributor to allow GRPC load balancing #148
+- [BUGFIX] Fixed `priorityClassName` in alertmanager deployment configuration. #155
+- [BUGFIX] Replacing deprecated (in Cortex 1.9) configuration with appropriate replacements (max_look_back_period -> max_query_lookback & compress_responses -> response_compression_enabled). #154
+- [BUGFIX] Added missing ServiceMonitors `additionalLabels` placeholders in `values.yaml`. #146
+- [ENHANCEMENT] Add Enable/Disable for security context #140
```
ShuzZzle, line 11: Buddy ....

 > the order of entries should be [CHANGE], [FEATURE], [ENHANCEMENT], [BUGFIX]
 
 Also did you really reformat the entire CHANGELOG.md? :laughing:

ShuzZzle (changes requested):
Also your forgot to DCO your commit

CHANGELOG.md
1995parham: Sorry, it was the prettifier and I have totally forgotten it. I will fix it now.

1995parham: About the reformat, revert it or it is okey?

ShuzZzle: the order matters but you don't have to fix the prettifier

1995parham: Thanks 👍

templates/alertmanager-dep.yaml
```diff
@@ -43,8 +43,9 @@ spec:
     {{- if .Values.alertmanager.priorityClassName }}
       priorityClassName: {{ .Values.alertmanager.priorityClassName }}
     {{- end }}
-      securityContext:
-        {{- toYaml .Values.alertmanager.securityContext | nindent 8 }}
+      {{- if .Values.alertmanager.securityContext.enabled }}
+      securityContext: {{- omit .Values.alertmanager.securityContext "enabled" | toYaml | nindent 8 }}
```
ShuzZzle, line 47: I just realised something, when you disable the securityContext for the **POD** then the securityContext is still active for the **container** as you can see in L86-88

ShuzZzle: Is that what we want?

1995parham: These are two different things I think and we can configure them separately. as you can see here we can have two different variables for them:

https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/statefulset.yaml

if you agree I can do this two but it introduces new variable.

ShuzZzle: Yeah that might be a good idea. Maybe refactor
```YAML
          securityContext:
            readOnlyRootFilesystem: true
```

into values.yml and do the same as above only with the containerSecurityContext variable. So I guess something amongst these lines

```YAML
[...]
      securityContext: {{- omit .Values.ruler.containerSecurityContext "enabled" | toYaml | nindent 8 }}
```
I think thats cool since we only have 1 container per "Pod" anyway. Otherwise that might a problem if you want to configure containers seperately

ShuzZzle: Or (probably the better idea) we could just call containerSecurityContext what is actually IS, i.e. rulerSecurityContext alertmanagersecurityContext ....

ShuzZzle: @cabrinha Whats your take on this

ShuzZzle:
Hi @1995parham 

We just merged #144. Due to the massive refactor you will have to rebase master and make sure that you don't actually add back the old files, since the location changed to a more easily maintainable folder structure

1995parham:
Yes, I will rebase it. about the permissions, they had execution permission and I removed it.

templates/alertmanager-dep.yaml
ShuzZzle: @1995parham Whats your opinion? Soon we will have more than 1 container in the deployment/set. When this happens it's not super clear anymore what containerSecurityContext actually does. Should we rename that to rulerSecurityContext so we can differentiate between multiple containers within in a deployment?

1995parham: I think we can prefix them with container names but also there is no harm if we use the same security context on all of them.
as an example, Bitnami charts do the same [here](https://github.com/bitnami/charts/blob/master/bitnami/kong/templates/dep-ds.yaml).

ShuzZzle (approved):
Reverted the CHANGELOG.md back to the original format. I think thats better but yeah LGTM!

ShuzZzle (approved):

1995parham:
Thanks for your changes and sorry about the Nginx security context.

PR #169, opened 2021-06-23
fix #168

### LINKED ISSUE #168: Ingester Ingested Samples Rate is doubled after upgrading to Cortex-helm-chart 0.5.0
I was using Cortex v1.7 and upgraded it to v1.9 yesterday. I used `cortex_distributor_received_samples_total` and `cortex_ingester_ingested_samples_total` to check distributor received rate and ingester ingested rate. And it was roughly same values when I'm running Cortex v1.7, but when I upgraded it to v1.9 `cortex_ingester_ingested_samples_total` metric value increased ~2.5x and `cortex_distributor_received_samples_total` value is staying same as before. I can see in-memory series(`cortex_ingester_memory_series`) in each ingester is also increased ~2.5x. Because of this ingester memory utilization is also higher than before. I have not increased the ingestion rate or did not configure new Prometheus remote writes, basically load is same as before. Anyone has any idea why this is happening ? 

I'm using below queries to get rates,
Distributor received samples rate : `sum(rate(cortex_distributor_received_samples_total[1m]))`
Ingester ingested samples rate : `sum(rate(cortex_ingester_ingested_samples_total[1m]))`

Below are the screenshots from Graphs before and after upgrade.

- Samples ingetion rates before,

alt text

- Samples ingetion rates after,

alt text

- In-memory series before,

alt text

- In-memory series after,

alt text

ShuzZzle:
**What this PR does**:
removed http-metrics port from every headless service and added ClusterIP service with http-metrics port to every component who before only had an headless service before. Headless services should not be used for anything http related, therefore I think its fine to completely remove that port from the Service.

No more dupe :) because of 2 Service with the same port
image

Fixes #168 

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 * [CHANGE] Removed dnssrvnoa resolution from block memcached (probably oversight) and moved back to simple dns resolution #164
-* [CHANGE] Removed http-metrics port from every headless service and added ClusterIP service with http-metrics port to every component who before only had an headless service before
+* [CHANGE] Removed http-metrics port from every headless service and added ClusterIP service with http-metrics port to every component who before only had an headless service before #169
 * [FEATURE] Support dynamic configuration of Ruler and AlertManager using sidecar #150
 * [ENHANCEMENT] Enable/Disable security & container security context #158
 * [ENHANCEMENT] ServiceMonitors: add options to configure metricRelabelings and relabelings #165
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 * [CHANGE] Removed dnssrvnoa resolution from block memcached (probably oversight) and moved back to simple dns resolution #164
+* [CHANGE] Removed http-metrics port from every headless service and added ClusterIP service with http-metrics port to every component who before only had an headless service before
 * [FEATURE] Support dynamic configuration of Ruler and AlertManager using sidecar #150
 * [ENHANCEMENT] Enable/Disable security & container security context #158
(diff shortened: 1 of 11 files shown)


templates/ingester/ingester-svc-headless.yaml
```diff
@@ -13,9 +13,9 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    - port: {{ .Values.config.server.http_listen_port }}
+    - port: {{ .Values.config.server.grpc_listen_port }}
       protocol: TCP
-      name: http-metrics
-      targetPort: http-metrics
+      name: grpc
+      targetPort: grpc
```
AndreZiviani, line 19: is this right?

ShuzZzle: there is a new ClusterIP Service which only does http-metrics, so basically I moved the http-metrics port to a new service

AndreZiviani: ok but the distributors were using the http port to connect to ingesters, with this they will use the grpc port, will it work without other changes?

ShuzZzle: @AndreZiviani where do you got that information? Communication between component is almost always (unless mistaken) exclusively grpc

ShuzZzle: Besides the Chart's included memberlist Service automatically selects every cortex component with the following label
`app.kubernetes.io/part-of: memberlist` and then the communication should appear through the 7946 port for memberlist which is exposed by the pods

ShuzZzle: But why don't you try it out and tell me if its working for you :)

AndreZiviani: Applied here and everything is still working :smile:

cabrinha (approved):

PR #182, opened 2021-07-28
add autoscaler for ingesters

kd7lxl:
**What this PR does**:
Adds a `HorizontalPodAutoscaler` for the ingesters.

https://cortexmetrics.io/docs/guides/capacity-planning/ describes that ingester memory grows as series are added, so this autoscaler tracks memory usage to make scaling decisions.

https://cortexmetrics.io/docs/guides/ingesters-scaling-up-and-down/#scaling-down says that "adding more ingesters to a Cortex cluster is considered a safe operation", so this autoscaler behavior is configured to allow adding one ingester pod at a time, but does not automatically scale down.

**Checklist**

### DIFF
--- a/templates/ingester/ingester-dep.yaml
+++ b/templates/ingester/ingester-dep.yaml
@@ -9,7 +9,9 @@ metadata:
   annotations:
     {{- toYaml .Values.ingester.annotations | nindent 4 }}
 spec:
+  {{- if not .Values.ingester.autoscaling.enabled }}
   replicas: {{ .Values.ingester.replicas }}
+  {{- end }}
   selector:
     matchLabels:
       {{- include "cortex.ingesterSelectorLabels" . | nindent 6 }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [FEATURE] Add autoscaler for ingesters #182
 * [ENHANCEMENT] Use FQDN for memcached addresses #175
 
 ## 0.6.0 / 2021-06-28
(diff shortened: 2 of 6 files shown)


templates/ingester/ingester-dep.yaml
```diff
@@ -9,7 +9,9 @@ metadata:
   annotations:
     {{- toYaml .Values.ingester.annotations | nindent 4 }}
 spec:
+  {{- if not .Values.ingester.autoscaling.enabled }}
```
ShuzZzle, line 12: Is this a good Idea? I think we should always have a minimum replica.

Does the Deployment even start without specifying replicas?

ShuzZzle, line 12: Is this a good Idea? I think we should always have a minimum replica.

Does the Deployment even start without specifying replicas?

kd7lxl: > Is this a good Idea? I think we should always have a minimum replica.

This is to handle https://github.com/kubernetes/kubernetes/issues/25238. `helm create` uses the same trick: https://github.com/helm/helm/blob/d4cc130fa90027e8c1c69ad2a06c6dee4c31d29d/pkg/chartutil/create.go#L285-L287 It's considered best practice.

> Does the Deployment even start without specifying replicas?

Yes, Deployments default to 1 replica when not specified (this behavior is documented in the ticket above). After creating the deployment, any patches to it will not affect the replicas; only the HPA will change the replicas field at that point. If you were to deploy the chart with this change and autoscaling enabled, the deployment would be created with 1 replica (default) and then moments later the HPA would increase it to the minimum of 3.

ShuzZzle: alright

kd7lxl:
Oops, you beat me to the merge/rebase.

ShuzZzle (approved):

PR #183, opened 2021-07-28
check for updated README.md in CI

kd7lxl:
**What this PR does**:
I noticed that the README.md had not been completely updated for the 0.6.0 release. This PR adds a PR check to make sure it has been updated.

**Checklist**

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -35,6 +35,11 @@ jobs:
       - name: Run chart-testing (lint)
         run: ct lint --config ct.yaml
 
+      - name: Updated README.md
+        run: |
+          make README.md
+          git diff --exit-code -- README.md
+
       - name: Create kind cluster
         uses: helm/kind-action@v1.2.0
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,3 @@
+.PHONY: README.md
+README.md:
+	docker run --rm --volume "$(shell pwd):/helm-docs" -u $(shell id -u) jnorwood/helm-docs:latest
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
+<!-- README.md is a generated file. Make any changes in README.md.gotmpl or values.yaml. -->
+
 # cortex
 
-![Version: 0.5.0](https://img.shields.io/badge/Version-0.5.0-informational?style=flat-square) ![AppVersion: v1.9.0](https://img.shields.io/badge/AppVersion-v1.9.0-informational?style=flat-square)
+![Version: 0.6.0](https://img.shields.io/badge/Version-0.6.0-informational?style=flat-square) ![AppVersion: v1.9.0](https://img.shields.io/badge/AppVersion-v1.9.0-informational?style=flat-square)
 
 Horizontally scalable, highly available, multi-tenant, long term Prometheus.
 
@@ -143,7 +145,7 @@ metadata:
   labels:
     cortex_alertmanager: "1"
   name: alertmanager-example-config
-data: 
+data:
   fake.yaml: |-
     global:
       resolve_timeout: 5m
@@ -296,9 +298,9 @@ Kubernetes: `^1.19.0-0`
 | compactor.strategy.type | string | `"RollingUpdate"` |  |
 | compactor.terminationGracePeriodSeconds | int | `240` |  |
 | compactor.tolerations | list | `[]` |  |
+| config.alertmanager.enable_api | bool | `false` | Enable the experimental alertmanager config api. |
 | config.alertmanager.external_url | string | `"/api/prom/alertmanager"` |  |
-| config.alertmanager.enable_api | bool | `false` | Enable the experimental alertmanager config api.  |
-| config.alertmanager.storage | object | {} | Type of backend to use to store alertmanager configs. Supported values are: "configdb", "gcs", "s3", "local". refer to: https://cortexmetrics.io/docs/configuration/configuration-file/#alertmanager_config |
+| config.alertmanager.storage | object | `{}` | Type of backend to use to store alertmanager configs. Supported values are: "configdb", "gcs", "s3", "local". refer to: https://cortexmetrics.io/docs/configuration/configuration-file/#alertmanager_config |
 | config.api.prometheus_http_prefix | string | `"/prometheus"` |  |
 | config.api.response_compression_enabled | bool | `true` |  |
 | config.auth_enabled | bool | `false` |  |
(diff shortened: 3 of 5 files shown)


ShuzZzle (approved):

ShuzZzle:
yeah sorry about that, Test should pass as soon as we merge #186.

.github/workflows/lint-test.yml
```diff
@@ -35,6 +35,11 @@ jobs:
       - name: Run chart-testing (lint)
         run: ct lint --config ct.yaml
 
+      - name: Updated README.md
+        run: |
+          make README.md
+          git diff --exit-code -- README.md
```
ShuzZzle, line 41: I think its a really good idea. But I am unsure how well that will work, when people try to **manually** add their flags to the helm option table

ShuzZzle: Maybe we should document the importance of using helm-docs while also making sure that people write their updates in the README.md.gotmpl

kd7lxl: In another project I maintain, the helm-docs job commits the rendered README.md directly to the PR instead of bailing out with `git diff --exit-code`. I didn't do that here because I figured people would need to manually update README.md.gotmp.

The first thing I encountered when opening a PR was this comment from the template:
> 1. Read our CONTRIBUTING.md guide

But there's no CONTRIBUTING.md! There's one in the main Cortex project, but it doesn't mention helm-docs. This repo probably needs its own CONTRIBUTING.md, and it could describe how to update README.md.gotmpl and README.md.  As a first-time contributor, I don't think I'm the best author for that.

ShuzZzle: I agree, the contributing stuff is completely missing unfortunately. I will spin something up in the next few days

ShuzZzle: > In another project I maintain, the helm-docs job commits the rendered README.md directly to the PR instead of bailing out with git diff --exit-code

Maybe we should just do this already.. and be done with the stupid manual editing. Do you have any Ideas on how to achieve that?

kd7lxl: > Do you have any Ideas on how to achieve that?

Here's the other project I did it in: https://github.com/infobloxopen/konk/commit/d1d5083efe0501aab4193344b6b761587ac215ab#diff-1eb4e5fd5611777d4e597ef299a1cb5ba8050c28a2dabbd4fbc56205d69e5ddaR160-R164

After running `helm-docs`, it runs the https://github.com/EndBug/add-and-commit action.

I can update my PR to add this if that's your preference. Yes?

ShuzZzle: Yeah that does seem like a good idea :)

kd7lxl: I added a workflow to generate README.md. Please approve running this workflow so we can see if it works.

.github/workflows/readme.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - name: Generate README.md
+        run: |
+          make README.md
+
+      - name: Commit changes
+        uses: EndBug/add-and-commit@v7
+        with:
+          message: generate README.md
+          add: README.md
```
ShuzZzle, line 27: maybe we should also set

```yaml
with:
    default_author: github_actor
    signoff: true
```
well atleast the signoff part should be necessarily since we check for that

kd7lxl: Added this.

kd7lxl:
It's working now (committing and pushing the README to the PR branch). To work on fork PRs, it runs on the `push` event in the fork's context. I'm not sure if this precludes it from reporting a PR status or if we don't see the PR status now simply because it hasn't been approved to run here.

ShuzZzle:
> It's working now (committing and pushing the README to the PR branch). To work on fork PRs, it runs on the `push` event in the fork's context. I'm not sure if this precludes it from reporting a PR status or if we don't see the PR status now simply because it hasn't been approved to run here.

Yeah something ain't right. There is nothing to approve on my end. Also before it showed 3 checks, now its just back to two. Is that expected? Maybe its bugged and we need to push an empty commit

ShuzZzle:
ah nice the good ol' restart. lmao haha

kd7lxl:
Oh, I know what that is. Github doesn't allow workflows to trigger other workflows (to avoid infinite loops), so the commit pushed by Github Actions will not trigger the Lint and Test workflow. This sucks for our use case.

ShuzZzle:
> Oh, I know what that is. Github doesn't allow workflows to trigger other workflows (to avoid infinite loops), so the commit pushed by Github Actions will not trigger the Lint and Test workflow. This sucks for our use case.

Okay I will hold-off from merging for now. Maybe we can use this: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run

And then configured the Lint / Test Flow to run only for helm chart changes and not for updated readmes. Because that shouldn't matter anyway. I know in GitLab CI you can ignore files.

kd7lxl:
> Okay I will hold-off from merging for now.

I can drop the part of the PR that generates and commits, returning it to how it was at first, if we don't have a good solution for triggering the tests from a generated commit.

ShuzZzle:
> > Okay I will hold-off from merging for now.
> 
> I can drop the part of the PR that generates and commits, returning it to how it was at first, if we don't have a good solution for triggering the tests from a generated commit.

You don't think simply ignoring these files will not work?

kd7lxl:
> You don't think simply ignoring these files will not work?

I don't think filtering by files is related. Github doesn't allow a commit from Github Actions to trigger another workflow, but to get it to show up in the list of PR statuses it has to be triggered by the commit. There might be a workaround for this, but it's not obvious to me and I think the other work on this PR (the makefile and PR test) can be helpful even without the automation.

ShuzZzle:
> > You don't think simply ignoring these files will not work?
> 
> I don't think filtering by files is related. Github doesn't allow a commit from Github Actions to trigger another workflow, but to get it to show up in the list of PR statuses it has to be triggered by the commit. There might be a workaround for this, but it's not obvious to me and I think the other work on this PR (the makefile and PR test) can be helpful even without the automation.

Yes but maybe we block the "triggering" in the first place if we simply ignore these files. We should do that anyway! No reason to run the complete pipeline when we simply push docs

kd7lxl:
Okay, I was assuming you wanted the Lint and Test workflow to be required on every PR, since that's how the branch protection is set up now. If you don't mind it not running on these commits, it's already doing that successfully (though accidentally) since Github isn't triggering it.

The caveat to only running tests when specific files are modified is that Github branch protection does not have a way to say "only require this status check to pass when specific files are modified", so you would need to remove the Lint and Test status checks from the branch protection settings if you are going to only run it for specific files.

Yes, if you don't mind iterative improvements, please merge this now and we'll work on v2 when we have a better idea of what if possible and what we want.

ShuzZzle:
> Yes, if you don't mind iterative improvements, please merge this now and we'll work on v2 when we have a better idea of what if possible and what we want.

No Problem :). Thank you for your time and the contribution

kd7lxl:
I opened #188 for the "generate readme" work and noted the problems with the draft solution.

PR #184, opened 2021-07-28
define namespace in templates

kd7lxl:
**What this PR does**:
This explicitly sets `.metadata.namespace` on resources based on the value supplied with `helm --namespace`. This is useful when deploying with tools that separate the render and deploy steps rather than use a simple `helm install ...`. Consider the simple example `helm template --namespace ... | kubectl apply -f -`; in this case all the cortex resources are deployed to the default namespace unless the template explicitly defines `.metadata.namespace`. A more complex variation of this example might archive the rendered manifest, and for auditing purposes it's helpful to have the namespace defined.
```
helm template --namespace ... > manifest.yaml
archive manifest.yaml
kubectl apply -f manifest.yaml
```

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master / unreleased
 
 * [FEATURE] Add autoscaler for ingesters #182
+* [ENHANCEMENT] Define namespace in templates #184
 * [ENHANCEMENT] Use FQDN for memcached addresses #175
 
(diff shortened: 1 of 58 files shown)


ShuzZzle:
Actually surprised that we dont actually do that yet. Good Idea

ShuzZzle (approved):

PR #181, opened 2021-07-28
Updated Readme for AlertManager & Ruler sidecar configuration

rgroemmer:
Signed-off-by: Raphael Grömmer <r.groemmer@yahoo.de>

I had a little bit of struggle configuring the sidecar for the `alertManager` & `ruler` with the actual documentation, after successfully accomplish the configuration I have some changes for the `README.md`.

**What this PR does**:
- Updated the storage configuration example
- Added a example `ConfigMap` containing a `alertmanager-config`
- Added some missing helm values to the values table.
 
**Checklist**
- [ x] `README.md` updated`

PR #186, opened 2021-07-29
update docs also in the tmpl file

ShuzZzle:
Signed-off-by: ShuzZzle <niclas.schad@gmail.com>

**What this PR does**:

Triggerhappy me forgot to update the changes in #181 also in the template. This pr should fix that

Persist #181

### DIFF
--- a/README.md.gotmpl
+++ b/README.md.gotmpl
@@ -88,18 +88,19 @@ alertmanager:
 ```
 And here are the related configuration values in AlertManager and Ruler:
 ```yaml
-alertmanager:
-  data_dir: /data/
-  storage:
-    type : local
-    local:
-      path: /data
-ruler:
-  rule_path: /data/rules
-  storage:
-    type : local
-    local:
-      directory: /tmp/rules
+config:
+  alertmanager:
+    data_dir: /data/
+    storage:
+      type : local
+      local:
+        path: /data
+  ruler:
+    rule_path: /data/rules
+    storage:
+      type : local
+      local:
+        directory: /tmp/rules
 ```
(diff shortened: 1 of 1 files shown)


kd7lxl:
Should also update values.yaml to document these options: https://github.com/cortexproject/cortex-helm-chart/pull/181/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R300

ShuzZzle:
> Should also update values.yaml to document these options: https://github.com/cortexproject/cortex-helm-chart/pull/181/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R300

Yeah that's exactly the Issue. The generated output often does not match the "human" input.

kd7lxl:
> Yeah that's exactly the Issue. The generated output often does not match the "human" input. We should just automate the whole process. Especially the releasing part of new versions

As a helm user, I often look first in values.yaml to see the available options rather than looking to the README.md. In this case, the values are not mentioned at all in the values.yaml. This breaks normal human usage as well as the helm-docs generator.

ShuzZzle:
> > Yeah that's exactly the Issue. The generated output often does not match the "human" input. We should just automate the whole process. Especially the releasing part of new versions
> 
> As a helm user, I often look first in values.yaml to see the available options rather than looking to the README.md. In this case, the values are not mentioned at all in the values.yaml. This breaks normal human usage as well as the helm-docs generator.

Absolutely. Sorry I misread your comment, I thought you meant the README.md (which I was talking about). We will definitely try to keep the values.yml as the single source of truth.

cabrinha (approved):

PR #188, opened 2021-07-30
add workflow to generate README.md

### LINKED ISSUE #187: Automatically generate/update README.md every PR
For ease of use, the README.md with the "helm options table" should be generated for each PR. We are currently using helm-docs.

@kd7lxl Had some good ideas on how to achieve that.

kd7lxl:
**What this PR does**:
This adds a Github Action workflow to automate running `make` (`helm-docs`) on each PR and committing any changes.

This is a continuation of the work in #183.

Github Actions workflows will not trigger on commits pushed by Github Actions (differentiated by the auto auth token), so this workflow will cause the Lint and Test workflow to not run. Since that is a required workflow, the PR cannot be merged until Lint and Test is triggered some other way, like closing and reopening the PR. I've left this PR in draft mode while we try to find the best solution to automate running `make` without defeating the required PR status checks.

Fixes #187

### DIFF
--- a/.github/workflows/readme.yml
+++ b/.github/workflows/readme.yml
@@ -1,0 +1,30 @@
+name: Generate README.md
+
+on:
+  push:
+    branches-ignore:
+      - main
+      - master
+    paths:
+      - values.yaml
+
+jobs:
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Generate README.md
+        run: |
+          make README.md
+
+      - name: Commit changes
+        uses: EndBug/add-and-commit@v7
+        with:
+          add: README.md
+          default_author: github_actions
+          message: "[skip ci] generate README.md"
+          signoff: true
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -2,9 +2,14 @@ name: Lint and Test Charts
 
 on:
   push:
-    branches:
+    branches-ignore:
+      - master
       - main
   pull_request:
+    paths-ignore:
+      - '*.md'
+      - 'docs/**'
+      - 'tools/**'
 
 jobs:
   lint-test:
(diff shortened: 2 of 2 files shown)


ShuzZzle:
We need just need to test it here but we might get away with 
```yaml
on:
  push:
    paths-ignore:
      - 'README.md*'
      - 'docs/**'
    branches:
      - main
      - master
  pull_request:
  ```
  
  and readme.yml workflow
  ```
  on:
  push:
    paths:
      - 'README.md*'
      - values.yaml
  pull_request:
```

Also I am not completely sure what happens when the ci test workflow is not run the 2nd time. Is the check just skipped? Is it marked as failed? Does the infinite-loop protection not even bother to check if it has to run the workflow (then it wouldn't work anyway)?

kd7lxl:
> Also I am not completely sure what happens when the ci test workflow is not run the 2nd time. Is the check just skipped?

Yes, it is just skipped, so it stays 🟠 indefinitely.

kd7lxl:
> ```yaml
> on:
>   push:
>     branches:
>       // kinda pointless, nobody should ever push to main/master
>       - main
>       - master
> ```

BTW, this would run any time a PR was merged to `main`/`master`. If it tries to commit, it will fail since branch protection is enabled.

ShuzZzle:
> > ```yaml
> > on:
> >   push:
> >     branches:
> >       // kinda pointless, nobody should ever push to main/master
> >       - main
> >       - master
> > ```
> 
> BTW, this would run any time a PR was merged to `main`/`master`. If it tries to commit, it will fail since branch protection is enabled.

yep thats why I added the comment. I am gonna ask some people, maybe somebody has a bright idea

.github/workflows/readme.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Generate README.md
+        run: |
+          make README.md
+
+      - name: Commit changes
+        uses: EndBug/add-and-commit@v7
+        with:
+          add: README.md
+          default_author: github_actions
+          message: "[skip ci] generate README.md"
```
ShuzZzle, line 29: this should fix our infinite-loop problem @kd7lxl

ShuzZzle (approved):

.github/workflows/readme.yml
kd7lxl: lgtm!

ShuzZzle:
Alright I think its fine currently how it is @kd7lxl 
Do you wanna give your final Approval or do you have any objections?

kd7lxl:
lgtm

PR #190, opened 2021-07-30
add autoscaler for queriers

kd7lxl:
**What this PR does**:
Adds a `HorizontalPodAutoscaler` for the queriers.

https://cortexmetrics.io/docs/architecture/#querier
> Queriers are stateless and can be scaled up and down as needed.

**Checklist**

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -819,6 +819,17 @@ querier:
             topologyKey: 'kubernetes.io/hostname'
 
   annotations: {}
+
+  autoscaling:
+    # -- Creates a HorizontalPodAutoscaler for the querier pods.
+    enabled: false
+    minReplicas: 2
+    maxReplicas: 30
+    targetCPUUtilizationPercentage: 80
+    targetMemoryUtilizationPercentage: 0  # 80
+    # -- Ref: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-configurable-scaling-behavior
+    behavior: {}
+
   persistence:
     subPath:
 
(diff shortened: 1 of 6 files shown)


values.yaml
```diff
@@ -819,6 +819,17 @@ querier:
...
 
   annotations: {}
+
+  autoscaling:
+    # -- Creates a HorizontalPodAutoscaler for the querier pods.
+    enabled: false
+    minReplicas: 2
+    maxReplicas: 30
+    targetCPUUtilizationPercentage: 80
+    targetMemoryUtilizationPercentage: 0  # 80
```
ShuzZzle, line 829: nice but i don't like the defaults. querier is also a highly dependant memory process, even though query frontend already splits querier by day.

I actually don't think CPU matters that much

ShuzZzle (changes requested):

ShuzZzle (approved):

values.yaml
ShuzZzle: @cabrinha what's your opinion?

kd7lxl: Somewhat related, I've spent a lot of time tuning resource definitions for each Cortex component and still don't feel like I've got it dialed in. I'd appreciate any real world knowledge people could share in the wiki or elsewhere.

ShuzZzle: >  still don't feel like I've got it dialed in

Yep its like a never ending story :)

cabrinha:
It'd also be nice if we could add HPA for every component that can safely be autoscaled.

kd7lxl:
> It'd also be nice if we could add HPA for every component that can safely be autoscaled.

I've been adding them one at a time as a learn/recognize that they are safe to autoscale, and by what metric. If someone with more operational experience could share appropriate scaling metrics for each component, I could add the autoscalers.

PR #193, opened 2021-08-04
add example values for block storage

kd7lxl:
**What this PR does**:
Provides example helm values to configure Cortex for blocks storage on AWS S3.

**Checklist**

### DIFF
--- a/conf/provider/aws-s3-values.yaml
+++ b/conf/provider/aws-s3-values.yaml
@@ -26,7 +26,7 @@ config:
     # format. This option should be set when using the blocks storage and the
     # store-gateway sharding is disabled (when enabled, the store-gateway instances
     # form a ring and addresses are picked from the ring).
-    store_gateway_addresses: cortex-store-gateway-headless.cortex.svc:9095
+    store_gateway_addresses: dns+cortex-store-gateway-headless.cortex.svc:9095
   ruler_storage:
     s3:
       bucket_name: # your bucket name
--- a/conf/provider/aws-s3-values.yaml
+++ b/conf/provider/aws-s3-values.yaml
@@ -1,0 +1,34 @@
+# This file provides helm values to configure Cortex for blocks storage on AWS S3.
+# Be sure to edit region, endpoint, and bucket_name for your environment.
+config:
+  alertmanager_storage:
+    s3:
+      bucket_name: # your bucket name
+      region: us-east-1
+      endpoint: s3.us-east-1.amazonaws.com
+  storage:
+    engine: blocks
+  blocks_storage:
+    backend: s3
+    s3:
+      bucket_name: # your bucket name
+      region: us-east-1
+      # -- The S3 bucket endpoint. It could be an AWS S3 endpoint listed at
+      # https://docs.aws.amazon.com/general/latest/gr/s3.html or the address of an
+      # S3-compatible service in hostname:port format.
+      endpoint: s3.us-east-1.amazonaws.com
+    bucket_store:
+      sync_dir: "/data"
+    tsdb:
+      dir: "/data"
+  querier:
+    # -- Comma separated list of store-gateway addresses in DNS Service Discovery
+    # format. This option should be set when using the blocks storage and the
+    # store-gateway sharding is disabled (when enabled, the store-gateway instances
+    # form a ring and addresses are picked from the ring).
+    store_gateway_addresses: cortex-store-gateway-headless.cortex.svc:9095
+  ruler_storage:
+    s3:
+      bucket_name: # your bucket name
+      region: us-east-1
+      endpoint: s3.us-east-1.amazonaws.com
--- a/README.md
+++ b/README.md
@@ -74,6 +74,11 @@ Note that it might be necessary to use `--reset-values` since some default value
 Source code can be found [here](https://cortexmetrics.io/)
 
 ## Usage
+
+### Examples
+
+Example values for various configurations can be found in [./conf](./conf)
+
 ### Rules and AlertManager configuration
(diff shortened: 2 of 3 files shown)


conf/provider/aws-s3-values.yaml
```diff
@@ -0,0 +1,34 @@
...
+    bucket_store:
+      sync_dir: "/data"
+    tsdb:
+      dir: "/data"
+  querier:
+    # -- Comma separated list of store-gateway addresses in DNS Service Discovery
+    # format. This option should be set when using the blocks storage and the
+    # store-gateway sharding is disabled (when enabled, the store-gateway instances
+    # form a ring and addresses are picked from the ring).
+    store_gateway_addresses: cortex-store-gateway-headless.cortex.svc:9095
```
kd7lxl, line 29: Is this supposed to be like? (It works for me as-is.)
```suggestion
    store_gateway_addresses: dns+cortex-store-gateway-headless.cortex.svc:9095
```

conf/provider/aws-s3-values.yaml
```diff
@@ -0,0 +1,34 @@
...
+      bucket_name: # your bucket name
+      region: us-east-1
+      # -- The S3 bucket endpoint. It could be an AWS S3 endpoint listed at
+      # https://docs.aws.amazon.com/general/latest/gr/s3.html or the address of an
+      # S3-compatible service in hostname:port format.
+      endpoint: s3.us-east-1.amazonaws.com
+    bucket_store:
+      sync_dir: "/data"
+    tsdb:
+      dir: "/data"
```
kd7lxl, line 23: These should probably be in the default values.yaml, since they need to match the path the emptyDir is mounted, which is statically defined in the deployments.

conf/provider/aws-s3-values.yaml
ShuzZzle: > Comma separated list of store-gateway addresses in DNS Service Discovery format
  
According to the docs yes.

conf/provider/aws-s3-values.yaml
ShuzZzle: What do have these in common with querier? The config you showed is for ingesters?

ShuzZzle:
@kd7lxl we could also use this https://github.com/cortexproject/cortex-helm-chart/wiki
That might actually be easier, no idea if you can create PR's for the Wiki

conf/provider/aws-s3-values.yaml
ShuzZzle: dns+ is most likely the default

conf/provider/aws-s3-values.yaml
kd7lxl: The docs say it's for the querier:
https://cortexmetrics.io/docs/configuration/configuration-file/#blocks_storage_config
> ```
> # This configures how the querier and store-gateway discover and synchronize
> # blocks stored in the bucket.
> bucket_store:
>   # Directory to store synchronized TSDB index headers.
>   # CLI flag: -blocks-storage.bucket-store.sync-dir
>   [sync_dir: <string> | default = "tsdb-sync"]
> ```

kd7lxl:
> @kd7lxl we could also use this https://github.com/cortexproject/cortex-helm-chart/wiki
> That might actually be easier, no idea if you can create PR's for the Wiki

The thing I don't like about the wiki is you don't end up with a file you can use directly with `helm install -f conf/provider/...`

ShuzZzle:
> > @kd7lxl we could also use this https://github.com/cortexproject/cortex-helm-chart/wiki
> > That might actually be easier, no idea if you can create PR's for the Wiki
> 
> The thing I don't like about the wiki is you don't end up with a file you can `cp` or use with `helm install -f conf/provider/...`

true but I don't think anybody does that

conf/provider/aws-s3-values.yaml
ShuzZzle: sync_dir is most definitely a folder in the ingester. Querier is a stateless process

kd7lxl: In that case, here is the relevant reference.
https://github.com/cortexproject/cortex-helm-chart/blob/5061ce897f279b9c55ead01b6a9cb163d8f74d6c/templates/ingester/ingester-dep.yaml#L72
The same point about configuring it in default values applies. The ingester should be configured to only write to directories that are mounted as volumes, not within the container. Should I open another PR to configure all write paths to be in volumes?

ShuzZzle: It already does that if you configure it correctly. ("/data/**/*")
Besides the security policy forbids anything to be written in the container root filesystem.
But if you want so set better defaults then I do not have any objections

kd7lxl:
> > > @kd7lxl we could also use this https://github.com/cortexproject/cortex-helm-chart/wiki
> > > That might actually be easier, no idea if you can create PR's for the Wiki
> > 
> > 
> > The thing I don't like about the wiki is you don't end up with a file you can `cp` or use with `helm install -f conf/provider/...`
> 
> true but I don't think anybody does that

I do.

Feel free to take this content as the basis for a wiki article and close the PR if you prefer.

conf/provider/aws-s3-values.yaml
kd7lxl: > if you configure it correctly.

This is the part I think is in scope for a helm chart to do by default. 

> Besides the security policy forbids anything to be written in the container root filesystem.

Oh, I missed this. What I might do then is add an e2e test that proves there are no errors with the default configuration and policy. 

I'll have to dig back through our source control to figure out why the operators of our instance did not find the defaults satisfactory.

ShuzZzle: >> if you configure it correctly.

> This is the part I think is in scope for a helm chart to do by default.

I agree. We should do that

ShuzZzle:
Added example config in the newly launched cortex wiki

PR #194, opened 2021-08-05
[ENHANCEMENT] Add a nginx config to authenticate requests to query-frontend svc (multi-tenancy)

arjunbhandage:
**What this PR does**: Adds a nginx config to authenticate requests to query-frontend svc when multi-tenancy is enabled. see https://cortexmetrics.io/docs/api/#authentication and https://cortexmetrics.io/docs/api/#querier--query-frontend

**Which issue(s) this PR fixes**: None

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [ENHANCEMENT] Define namespace in templates #184
 * [ENHANCEMENT] Use FQDN for memcached addresses #175
 * [ENHANCEMENT] Optionally generate endpoints for `X-Scope-OrgID` injection (multi-tenancy) #180
+* [ENHANCEMENT] Add a nginx config to authenticate requests to query-frontend svc (multi-tenancy) #194
 * [BUGFIX] Correcting nginx config for auth orgs to right proxy_pass #192
 
 ## 0.6.0 / 2021-06-28
--- a/templates/nginx/nginx-config.yaml
+++ b/templates/nginx/nginx-config.yaml
@@ -109,6 +109,11 @@ data:
           proxy_set_header      X-Scope-OrgID {{ $org }};
           proxy_pass      http://{{ template "cortex.fullname" $ }}-distributor.{{ $rootDomain }}/api/v1/push;
         }
+        location ~ ^{{$.Values.config.api.prometheus_http_prefix}}/api/v1/{{ $org }}/(read|metadata|labels|series|query_range|query) {
+          proxy_set_header      X-Scope-OrgID {{ $org }};
+          rewrite ^(/{{$.Values.config.api.prometheus_http_prefix}}/.*)/{{ $org }}/(\w+)\.?.*$ $1/$2 last;
+          proxy_pass      http://{{ template "cortex.fullname" $ }}-query-frontend.{{ $rootDomain }}$request_uri;
+        }
         {{- end }}
         {{- end }}
       }


arjunbhandage:
Need help here. The nginx config seems correct to me, but does not work when I deploy these changes. Followed the below blog for
https://www.nginx.com/blog/creating-nginx-rewrite-rules/
http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite

> Here’s a sample NGINX rewrite rule that uses the rewrite directive. It matches URLs that begin with the string /download and then include the /media/ or /audio/ directory somewhere later in the path. It replaces those elements with /mp3/ and adds the appropriate file extension, .mp3 or .ra. The $1 and $2 variables capture the path elements that aren't changing. As an example, /download/cdn-west/media/file1 becomes /download/cdn-west/mp3/file1.mp3. If there is an extension on the filename (such as .flv), the expression strips it off and replaces it with .mp3.

ShuzZzle:
> Need help here. The nginx config seems correct to me, but does not work when I deploy these changes. Followed the below blog for
> https://www.nginx.com/blog/creating-nginx-rewrite-rules/
> http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
> 
> > Here’s a sample NGINX rewrite rule that uses the rewrite directive. It matches URLs that begin with the string /download and then include the /media/ or /audio/ directory somewhere later in the path. It replaces those elements with /mp3/ and adds the appropriate file extension, .mp3 or .ra. The $1 and $2 variables capture the path elements that aren't changing. As an example, /download/cdn-west/media/file1 becomes /download/cdn-west/mp3/file1.mp3. If there is an extension on the filename (such as .flv), the expression strips it off and replaces it with .mp3.

Any errors in the nginx pod?

ShuzZzle:
> Need help here. The nginx config seems correct to me, but does not work when I deploy these changes. Followed the below blog for
> https://www.nginx.com/blog/creating-nginx-rewrite-rules/
> http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
> 
> > Here’s a sample NGINX rewrite rule that uses the rewrite directive. It matches URLs that begin with the string /download and then include the /media/ or /audio/ directory somewhere later in the path. It replaces those elements with /mp3/ and adds the appropriate file extension, .mp3 or .ra. The $1 and $2 variables capture the path elements that aren't changing. As an example, /download/cdn-west/media/file1 becomes /download/cdn-west/mp3/file1.mp3. If there is an extension on the filename (such as .flv), the expression strips it off and replaces it with .mp3.

Any error in the nginx pod?

arjunbhandage:
> > Need help here. The nginx config seems correct to me, but does not work when I deploy these changes. Followed the below blog for
> > https://www.nginx.com/blog/creating-nginx-rewrite-rules/
> > http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
> > > Here’s a sample NGINX rewrite rule that uses the rewrite directive. It matches URLs that begin with the string /download and then include the /media/ or /audio/ directory somewhere later in the path. It replaces those elements with /mp3/ and adds the appropriate file extension, .mp3 or .ra. The $1 and $2 variables capture the path elements that aren't changing. As an example, /download/cdn-west/media/file1 becomes /download/cdn-west/mp3/file1.mp3. If there is an extension on the filename (such as .flv), the expression strips it off and replaces it with .mp3.
> 
> Any error in the nginx pod?

None. I find it really difficult to debug this. What tried is exec'd into the nginx pod. and ran the below command.

What I get:
```
# curl -i http://localhost:80/prometheus/api/v1/org/read
HTTP/1.1 404 Not Found
Server: nginx/1.21.1
Date: Thu, 05 Aug 2021 10:37:30 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 19
Connection: keep-alive
X-Content-Type-Options: nosniff
```

What I expect from the command :
```
HTTP/1.1 405 Method Not Allowed
Server: nginx/1.21.1
Date: Thu, 05 Aug 2021 10:38:28 GMT
Content-Length: 0
Connection: keep-alive
```

arjunbhandage:
Not really required as we can configure the grafana data source with custom headers for authentication.
See https://grafana.com/docs/grafana/latest/administration/provisioning/#datasources
Custom HTTP headers for datasources

PR #195, opened 2021-08-06
graceful shutdown of ingesters

kd7lxl:
**What this PR does**:
https://cortexmetrics.io/docs/api/#shutdown says that `/ingester/shutdown` should be called before sending SIGTERM to an ingester. This PR adds a preStop hook to do that. This gives the ingester an opportunity to safely flush blocks from memory before being terminated.

#### Reference:
* https://cortexmetrics.io/docs/api/#shutdown
* https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks
* https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#lifecycle-v1-core

**Checklist**

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -537,9 +537,14 @@ ingester:
     targetMemoryUtilizationPercentage: 80
     behavior:
       scaleDown:
-        # -- Scaledown procedure varies, so automatic scaledown is disabled
-        # Ref: https://cortexmetrics.io/docs/guides/ingesters-scaling-up-and-down/#scaling-down
-        selectPolicy: Disabled
+        # -- see https://cortexmetrics.io/docs/guides/ingesters-scaling-up-and-down/#scaling-down for scaledown details
+        policies:
+          - type: Pods
+            value: 1
+            # set to no less than 2x the maximum between -blocks-storage.bucket-store.sync-interval and -compactor.cleanup-interval
+            periodSeconds: 1800
+        # -- uses metrics from the past 1h to make scaleDown decisions
+        stabilizationWindowSeconds: 3600
       scaleUp:
         # -- This default scaleup policy allows adding 1 pod every 30 minutes.
         # Ref: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-configurable-scaling-behavior
--- a/values.yaml
+++ b/values.yaml
@@ -553,6 +553,17 @@ ingester:
             value: 1
             periodSeconds: 1800
 
+  lifecycle:
+    # -- The /shutdown preStop hook is recommended as part of the ingester
+    # scaledown process, but can be removed to optimize rolling restarts in
+    # instances that will never be scaled down or when using chunks storage
+    # with WAL disabled.
+    # https://cortexmetrics.io/docs/guides/ingesters-scaling-up-and-down/#scaling-down
+    preStop:
+      httpGet:
+        path: "/ingester/shutdown"
+        port: http-metrics
+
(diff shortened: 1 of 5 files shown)


rsteneteg:
The docs says that if you send the /ingester/shutdown api call, it is expected user also send the the ingester a SIGTERM as that is not done automatically.

It does not say that you must send the /ingester/shutdown call before a SIGTERM, as the ingester will automatically do a graceful shutdown.

In some cases like if you run the Ingesters with WAL enabled, you may not want to call the shutdown api call to flush data to storage, since Ingesters with WAL will write in-memory data to the WAL on shutdown and read that data from the WAL on startup.

kd7lxl:
> It does not say that you must send the /ingester/shutdown call before a SIGTERM, as the ingester will automatically do a graceful shutdown.

It's here under step 1: https://cortexmetrics.io/docs/guides/ingesters-scaling-up-and-down/#blocks-storage

rsteneteg:
ok, you are right if you are doing a scale down where you want to reduce the number of Ingesters.

Does this change only apply to scale downs, or would this trigger a flush to storage on every shutdown during a rolling restart of the ingesters during upgrades etc?

kd7lxl:
> Does this change only apply to scale downs, or would this trigger a flush to storage on every shutdown during a rolling restart of the ingesters during upgrades etc?

This applies to everything -- Kubernetes doesn't differentiate.

What will happen in Cortex if this runs on every pod shutdown? Does the compactor know how to handle duplicate data that has been flushed to a block already, but also exists in WAL?

rsteneteg:
I am unsure about the duplication, but I am also considering how much longer rolling restarts would take, and query performance as the metrics needs to be downloaded from storage again.

In my use case we rarely need to scale down Ingesters, it is more important to scale up, and scale downs can be done manually if needed. Because of this I am wondering if we could put a toggle flag on this so users of the chart can choose if they want to flush on shutdown or not?

kd7lxl:
> I am unsure about the duplication, but I am also considering how much longer rolling restarts would take, and query performance as the metrics needs to be downloaded from storage again.

In my testing with ingesters sized to 4GB, pod termination with flushing took about 17 seconds longer.

> In my use case we rarely need to scale down Ingesters, it is more important to scale up, and scale downs can be done manually if needed. Because of this I am wondering if we could put a toggle flag on this so users of the chart can choose if they want to flush on shutdown or not?

It's not something we do often, but if we do want to scale down ingesters, we want it to happen gracefully and you can't modify the lifecycle hooks of an already-running pod so it has to be there to start with.

I can make it configurable...

values.yaml
```diff
@@ -553,6 +553,17 @@ ingester:
             value: 1
             periodSeconds: 1800
 
+  lifecycle:
+    # -- The /shutdown preStop hook is recommended as part of the ingester
+    # scaledown process, but can be removed to optimize rolling restarts in
+    # instances that will never be scaled down or when using chunks storage
+    # with WAL disabled.
```
ShuzZzle, line 560: ok but soon to be deprecated

ShuzZzle:
Nice just as a quick reminder

> The Pod's termination grace period countdown begins before the PreStop hook is executed, so regardless of the outcome of the handler, the container will eventually terminate within the Pod's termination grace period

Its a cool feature, but it might be pretty difficult to get it right. Per default we allow the termination grace period to be 240s. So the end user must ensure that the pod actually has enough time to flush, otherwise we might risk running into problems when the pod eventually is forcibly terminated

ShuzZzle (approved):

PR #197, opened 2021-08-12
exercise memcached in test

kd7lxl:
**What this PR does**:
exercise memcached features in chart test

### DIFF
--- a/ci/test-values.yaml
+++ b/ci/test-values.yaml
@@ -49,7 +49,7 @@ config:
       dir: "/data/store"
 
   querier:
-    store_gateway_addresses: "dns+cortex-store-gateway-headless:9095"
+    store_gateway_addresses: "dns+{{.Release.Name}}-store-gateway-headless:9095"
 
   memberlist:
     join_members:
--- a/ci/test-values.yaml
+++ b/ci/test-values.yaml
@@ -79,31 +79,53 @@ memcached:
   enabled: true
   architecture: standalone
   replicaCount: 1
-  resources: {}
+  resources:
+    # https://github.com/helm/helm/issues/9136
+    requests:
+      cpu: 0
+      memory: 0
 memcached-index-read:
   enabled: true
   architecture: standalone
   replicaCount: 1
-  resources: {}
+  resources:
+    requests:
+      cpu: 0
+      memory: 0
 memcached-index-write:
   enabled: true
   architecture: standalone
   replicaCount: 1
-  resources: {}
+  resources:
+    requests:
+      cpu: 0
+      memory: 0
 memcached-frontend:
   enabled: true
   architecture: standalone
   replicaCount: 1
-  resources: {}
+  resources:
+    requests:
+      cpu: 0
+      memory: 0
 memcached-blocks:
   architecture: standalone
   replicaCount: 1
-  resources: {}
+  resources:
+    requests:
+      cpu: 0
+      memory: 0
 memcached-blocks-index:
   architecture: standalone
   replicaCount: 1
-  resources: {}
+  resources:
+    requests:
+      cpu: 0
+      memory: 0
 memcached-blocks-metadata:
   architecture: standalone
   replicaCount: 1
-  resources: {}
+  resources:
+    requests:
+      cpu: 0
+      memory: 0
--- a/ci/test-values.yaml
+++ b/ci/test-values.yaml
@@ -49,11 +49,11 @@ config:
       dir: "/data/store"
 
   querier:
-    store_gateway_addresses: "dns+{{.Release.Name}}-store-gateway-headless:9095"
+    store_gateway_addresses: 'dns+{{ include "cortex.storeGatewayFullname" $ }}-headless:9095'
 
   memberlist:
     join_members:
-      - "{{.Release.Name}}-memberlist.cortex"
+      - '{{ include "cortex.fullname" $ }}-memberlist'
 
 ingester:
   replicas: 1
--- a/ci/test-values.yaml
+++ b/ci/test-values.yaml
@@ -79,24 +79,31 @@ memcached:
   enabled: true
   architecture: standalone
   replicaCount: 1
+  resources: {}
 memcached-index-read:
   enabled: true
   architecture: standalone
   replicaCount: 1
+  resources: {}
 memcached-index-write:
   enabled: true
   architecture: standalone
   replicaCount: 1
+  resources: {}
 memcached-frontend:
   enabled: true
   architecture: standalone
   replicaCount: 1
+  resources: {}
 memcached-blocks:
   architecture: standalone
   replicaCount: 1
+  resources: {}
 memcached-blocks-index:
   architecture: standalone
   replicaCount: 1
+  resources: {}
 memcached-blocks-metadata:
   architecture: standalone
   replicaCount: 1
+  resources: {}
--- a/ci/test-values.yaml
+++ b/ci/test-values.yaml
@@ -72,3 +72,31 @@ querier:
     enabled: true
 nginx:
   replicas: 1
+
+tags:
+  blocks-storage-memcached: true
+memcached:
+  enabled: true
+  architecture: standalone
+  replicaCount: 1
+memcached-index-read:
+  enabled: true
+  architecture: standalone
+  replicaCount: 1
+memcached-index-write:
+  enabled: true
+  architecture: standalone
+  replicaCount: 1
+memcached-frontend:
+  enabled: true
+  architecture: standalone
+  replicaCount: 1
+memcached-blocks:
+  architecture: standalone
+  replicaCount: 1
+memcached-blocks-index:
+  architecture: standalone
+  replicaCount: 1
+memcached-blocks-metadata:
+  architecture: standalone
+  replicaCount: 1
--- a/ci/test-values.yaml
+++ b/ci/test-values.yaml
@@ -53,7 +53,7 @@ config:
 
   memberlist:
     join_members:
-    - "cortex-memberlist.cortex"
+      - "{{.Release.Name}}-memberlist.cortex"
 
 ingester:
   replicas: 1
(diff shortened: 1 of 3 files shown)


kd7lxl:
https://github.com/cortexproject/cortex-helm-chart/blob/5061ce897f279b9c55ead01b6a9cb163d8f74d6c/ci/test-values.yaml#L54-L56

Test is failing because this is invalid. `ct` uses a random release name, e.g., `cortex-wou93c6h5x-memberlist` that doesn't match this config.

ShuzZzle:
> https://github.com/cortexproject/cortex-helm-chart/blob/5061ce897f279b9c55ead01b6a9cb163d8f74d6c/ci/test-values.yaml#L54-L56
> 
> Test is failing because this is invalid. `ct` uses a random release name, e.g., `cortex-wou93c6h5x-memberlist` that doesn't match this config.

Is there no way to disable the random suffix for releases? According to the code: https://github.com/helm/chart-testing/blob/main/pkg/chart/chart.go#L202 there is no way to disable this behaviour

kd7lxl:
The best solution is probably to have the chart build a template for the service name with `{{.Release.Name}}`, like it does for the memcached config. This way chart users aren't required to construct the value themselves.

ShuzZzle:
> The best solution is probably to have the chart build a template for the service name with `{{.Release.Name}}`, like it does for the memcached config. This way chart users aren't required to construct the value themselves.

I am afraid I can't quite follow. `.Release.Name` is already used across the cortex-helm-chart. The problem is that ct randomizes the release name at runtime and there is no real way of knowing that beforehand, unless we hardcode the headless memberlist service name so that it doesn't change anymore.

kd7lxl:
I'm saying the fix is to define

```yaml
 54 to 56 in 5061ce8

 memberlist: 
   join_members: 
   - "{{.Release.Name}}-memberlist.cortex" 
```
in a context that evaluates templates as opposed to in the ci values as it is now https://github.com/cortexproject/cortex-helm-chart/blob/5061ce897f279b9c55ead01b6a9cb163d8f74d6c/ci/test-values.yaml#L54-L56

ShuzZzle:
> I'm saying the fix is to define
> 
> ```yaml
>  54 to 56 in 5061ce8
> 
>  memberlist: 
>    join_members: 
>    - "{{.Release.Name}}-memberlist.cortex" 
> ```
> in a context that evaluates templates as opposed to in the ci values as it is now https://github.com/cortexproject/cortex-helm-chart/blob/5061ce897f279b9c55ead01b6a9cb163d8f74d6c/ci/test-values.yaml#L54-L56 

Oh yeah, of cause. I totally forget about that possibility. I like it

ShuzZzle:
I ran 
```
kubectl create namespace cortex
ct install --config ct.yaml --namespace cortex
```
against my kubernetes (1.20.9) cluster and there a quite a lot of issues. However it passes and finishes "successfull".

I ran the same again using
```
helm install cortex . --namespace=cortex --create-namespace -f ci/test-values.yaml
```
got the same errors but everything went "online" eventually, which brings me to my point. It just takes a while to get everything up and running, especially if we publish endpoints which are not ready yet. @kd7lxl Could it be that we just have to increase the timeout?

By the way the HPA doesn't really work for the ci since the querier doesn't have any resource/requests applied
> invalid metrics (1 invalid out of 1), first error is: failed to get cpu utilization: missing request for cpu
  Warning  FailedGetResourceMetric       3m26s (x40 over 13m)  horizontal-pod-autoscaler  failed to get cpu utilization: missing request for cpu


I can't find anything else

kd7lxl:
> I ran
> 
> ```
> kubectl create namespace cortex
> ct install --config ct.yaml --namespace cortex
> ```
> 
> against my kubernetes (1.20.9) cluster and there a quite a lot of issues. However it passes and finishes "successfully".
> 
> I ran the same again using
> 
> ```
> helm install cortex . --namespace=cortex --create-namespace -f ci/test-values.yaml
> ```
> 
> got the same errors but everything went "online" eventually, which brings me to my point. It just takes a while to get everything up and running, especially if we don't publish endpoints (`publishNotReadyAddresse` only for the distributor, query-frontend, alertmanager) which are not ready yet. @kd7lxl Could it be that we just have to increase the timeout?
> 
> I also checked the cortex status page and everything was healthy `¯\_(ツ)_/¯`

The problem I assumed we needed to solve was this, from the CI logs:
```
ts=2021-08-17T14:40:34.70607837Z caller=memberlist_logger.go:74 level=warn msg="Failed to resolve cortex-memberlist.cortex: lookup cortex-memberlist.cortex on 10.96.0.10:53: no such host"
```

But now with your change it looks like this, so I guess it's a problem with the memberlist endpoints not being ready immediately as opposed to the wrong hostname:
```
ts=2021-08-17T14:40:34.70607837Z caller=memberlist_logger.go:74 level=warn msg="Failed to resolve cortex-3p6brmjpl5-memberlist.cortex: lookup cortex-3p6brmjpl5-memberlist.cortex on 10.96.0.10:53: no such host"
```

This seems to clear up eventually. 👍 

> By the way the HPA doesn't really work for the ci since the querier doesn't have any resource/requests applied
> 
> > invalid metrics (1 invalid out of 1), first error is: failed to get cpu utilization: missing request for cpu
> > Warning  FailedGetResourceMetric       3m26s (x40 over 13m)  horizontal-pod-autoscaler  failed to get cpu utilization: missing request for cpu

Also, the metrics-server is not deployed in Kind. That's okay -- we don't need the autoscaler to scale in the test, but at least submitting the resource to k8s allows it to be checked for compliance.

> I can't find anything else
> 
> Noticeable Job output:
> 
> > W0817 14:40:02.390348    7312 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
> > W0817 14:40:02.392074    7312 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
> > W0817 14:40:02.561890    7312 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
> > W0817 14:40:02.563516    7312 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
> > **Error: timed out waiting for the condition**

We'll want to do something like this to handle the changing apiVersion: https://github.com/helm/helm/blob/d4cc130fa90027e8c1c69ad2a06c6dee4c31d29d/pkg/chartutil/create.go#L223-L229

ShuzZzle:
> > I can't find anything else
> > Noticeable Job output:
> > > W0817 14:40:02.390348    7312 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
> > > W0817 14:40:02.392074    7312 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
> > > W0817 14:40:02.561890    7312 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
> > > W0817 14:40:02.563516    7312 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
> > > **Error: timed out waiting for the condition**
> 
> We'll want to do something like this to handle the changing apiVersion: https://github.com/helm/helm/blob/d4cc130fa90027e8c1c69ad2a06c6dee4c31d29d/pkg/chartutil/create.go#L223-L229

Yeah we can control that with capabilities but I actually meant the bold marked line. I think the current problem is that it simply doesn't get ready in time. Which does make sense considering the extra pods, and that it aborts after roughly 10minutes. We can adjust the timeout in ct.yaml and increase it to 15min.

```YAML
helm-extra-args: --timeout 900s
```

ShuzZzle:
@kd7lxl  Alright found the culprit. Some memcached Pods can't be scheduler, not enough resources. Memcached comes with default requests and limits.

>   Warning  FailedScheduling  29s (x17 over 15m)  default-scheduler  0/1 nodes are available: 1 Insufficient cpu.

ci/test-values.yaml
```diff
@@ -49,11 +49,11 @@ config:
       dir: "/data/store"
 
   querier:
-    store_gateway_addresses: "dns+cortex-store-gateway-headless:9095"
+    store_gateway_addresses: "dns+{{.Release.Name}}-store-gateway-headless:9095"
 
   memberlist:
     join_members:
-    - "cortex-memberlist.cortex"
+      - "{{.Release.Name}}-memberlist.cortex"
```
kd7lxl, line 56: Neat, I hadn't noticed that the chart already evaluated the `.Values.config` as a template. This is perfect.

ci/test-values.yaml
```diff
@@ -72,3 +72,38 @@ querier:
...
 nginx:
   replicas: 1
+
+tags:
+  blocks-storage-memcached: true
+memcached:
+  enabled: true
+  architecture: standalone
+  replicaCount: 1
+  resources: {}
```
ShuzZzle, line 82: This `{}` doesn't really work for me. I think I am running into (https://github.com/helm/helm/issues/9136). I have also tried nullying the specific keys but it's not doing anything. Can you confirm?

ShuzZzle: Yikes how is that still a thing

ShuzZzle (approved):
About setting everything to 0:
I absolutely hate the solution but nothing really else to be done here.

-.-

ShuzZzle (approved):
lgtm

PR #199, opened 2021-08-17
fix alertmanager cluster

### LINKED ISSUE #174: Wrong Alertmanager Cluster SRV Record in the ruler
For Alertmanager auto discovery the url to the alertmanager cluster is wrong

Check: https://github.com/cortexproject/cortex-helm-chart/blob/v0.6.0/templates/ruler/ruler-dep.yaml#L110

the port is wrong and should be replaced by port "cluster". Should look like this
```YAML
- "-ruler.alertmanager-url=http://_cluster._tcp.{{ template "cortex.name" . }}-alertmanager-headless/api/prom/alertmanager/"
```
(Untested)

Also Check if that is used somewhere else

ShuzZzle:
**What this PR does**:
Tries to fix the alertmanager cluster feature. The new configuration looks correct according to the docs but further testing has to be done to confirm that it is behaving as expected. Also this is incompatible with cortex's sharding feature, you can't have both.

Fixes #174 

**Checklist**

### DIFF
--- a/templates/alertmanager/_helpers-alertmanager.tpl
+++ b/templates/alertmanager/_helpers-alertmanager.tpl
@@ -21,3 +21,17 @@ alertmanager selector labels
 {{ include "cortex.selectorLabels" . }}
 app.kubernetes.io/component: alertmanager
 {{- end }}
+
+{{/*
+alertmanager cluster peer list
+*/}}
+{{- define "cortex.alertmanagerClusterPeerList" -}}
+{{- $clusterPort := regexReplaceAll ".+[:]" (default "0.0.0.0:9094" .Values.config.alertmanager.cluster_bind_address) "" -}}
+{{- $alertClusterPeerList := dict "peers" (list) -}}
+{{- $cortexFullName := include "cortex.fullname" . -}}
+{{- $alertmanagerFullName := include "cortex.alertmanagerFullname" . -}}
+{{- range $n := until (.Values.alertmanager.replicas | int) }}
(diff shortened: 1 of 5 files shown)


templates/alertmanager/_helpers-alertmanager.tpl
```diff
@@ -21,3 +21,17 @@ alertmanager selector labels
...
+
+{{/*
+alertmanager cluster peer list
+*/}}
+{{- define "cortex.alertmanagerClusterPeerList" -}}
+{{- $clusterPort := regexReplaceAll ".+[:]" (default "0.0.0.0:9094" .Values.config.alertmanager.cluster_bind_address) "" -}}
+{{- $alertClusterPeerList := dict "peers" (list) -}}
+{{- $cortexFullName := include "cortex.fullname" . -}}
+{{- $alertmanagerFullName := include "cortex.alertmanagerFullname" . -}}
+{{- range $n := until (.Values.alertmanager.replicas | int) }}
```
ShuzZzle, line 33: I guess if we ever create an HPA for the alertmanager, then this won't really work

rastislavs:
> Wont-Do. Its considered legacy and we will not support as nobody uses it

@nschad Just run into the same issue. If Alertmanager cluster is considered legacy, is there any other way for providing HA for Alertmanager in Cortex?

nschad:
> > Wont-Do. Its considered legacy and we will not support as nobody uses it
> 
> @nschad Just run into the same issue. If Alertmanager cluster is considered legacy, is there any other way for providing HA for Alertmanager in Cortex?

You probably want 
```
# Shard tenants across multiple alertmanager instances.
# CLI flag: -alertmanager.sharding-enabled
[sharding_enabled: <boolean> | default = false]
```

PR #206, opened 2021-08-30
Add basic auth configuration options to Nginx

### LINKED ISSUE #205: Add Basic Auth to Nginx Config
Having the ability to configure the Nginx config, particularly the ability to configure basic authentication, is very desirable. 

Having something as simple as the following in the Nginx ConfigMap template would be great:

```
       {{ if .Values.nginx.config.basicAuthMountPath -}}
        auth_basic "Restricted Content";
        auth_basic_user_file {{ .Values.nginx.config.basicAuthMountPath }};
        {{- end }}
```
A similar issues is being discussed here: https://github.com/cortexproject/cortex-helm-chart/issues/72

logston:
**What this PR does**:
  This commit adds the ability to include an `auth_basic` and
  `auth_basic_user_file` directive in the Nginx configuration file.

Fixes #205 

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,6 +588,7 @@ Kubernetes: `^1.19.0-0`
 | nginx.&ZeroWidthSpace;affinity | object | `{}` |  |
 | nginx.&ZeroWidthSpace;annotations | object | `{}` |  |
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;auth_orgs | list | `[]` | (optional) List of [auth tenants](https://cortexmetrics.io/docs/guides/auth/) to set in the nginx config |
+| nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;basicAuthSecretName | string | `""` | (optional) Name of basic auth secret. In order to use this option, a secret with htpasswd formatted contents at the key ".htpasswd" must exist. For example:   apiVersion: v1   kind: Secret   metadata:     name: my-secret     namespace: <same as cortex installation>   stringData:     .htpasswd: |       user1:$apr1$/woC1jnP$KAh0SsVn5qeSMjTtn0E9Q0       user2:$apr1$QdR8fNLT$vbCEEzDj7LyqCMyNpSoBh/ Please note that the use of basic auth will not identify organizations the way X-Scope-OrgID does. Thus, the use of basic auth alone will not prevent one tenant from viewing the metrics of another. To ensure tenants are scoped appropriately, explicitly set the `X-Scope-OrgID` header in the nginx config. Example   setHeaders:     X-Scope-Org-Id: $remote_user |
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;client_max_body_size | string | `"1M"` |  |
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;dnsResolver | string | `"kube-dns.kube-system.svc.cluster.local"` |  |
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;setHeaders | object | `{}` |  |
--- a/templates/nginx/nginx-config.yaml
+++ b/templates/nginx/nginx-config.yaml
@@ -40,7 +40,18 @@ data:
         proxy_set_header {{ $key }} {{ $value }};
         {{- end }}
 
+        {{ if .Values.nginx.config.basicAuthSecretName -}}
+        auth_basic "Restricted Content";
+        auth_basic_user_file /etc/apache2/.htpasswd;
+        {{- end }}
+
         location = /healthz {
+          # auth_basic off is not set here, even when a basic auth directive is
+          # included in the server block, as Nginx's NGX_HTTP_REWRITE_PHASE
+          # (point when this return statement is evaluated) comes before the
+          # NGX_HTTP_ACCESS_PHASE (point when basic auth is evaluated). Thus,
+          # this return statement returns a response before basic auth is
+          # evaluated.
           return 200 'alive';
         }
 
--- a/values.yaml
+++ b/values.yaml
@@ -1165,6 +1165,29 @@ nginx:
     setHeaders: {}
     # -- (optional) List of [auth tenants](https://cortexmetrics.io/docs/guides/auth/) to set in the nginx config
     auth_orgs: []
+    # -- (optional) Name of basic auth secret.
+    # In order to use this option, a secret with htpasswd formatted contents at
+    # the key ".htpasswd" must exist. For example:
+    #
+    #   apiVersion: v1
+    #   kind: Secret
+    #   metadata:
+    #     name: my-secret
+    #     namespace: <same as cortex installation>
+    #   stringData:
+    #     .htpasswd: |
+    #       user1:$apr1$/woC1jnP$KAh0SsVn5qeSMjTtn0E9Q0
+    #       user2:$apr1$QdR8fNLT$vbCEEzDj7LyqCMyNpSoBh/
+    #
+    # Please note that the use of basic auth will not identify organizations
+    # the way X-Scope-OrgID does. Thus, the use of basic auth alone will not
+    # prevent one tenant from viewing the metrics of another. To ensure tenants
+    # are scoped appropriately, explicitly set the `X-Scope-OrgID` header
+    # in the nginx config. Example
+    #   setHeaders:
+    #     X-Scope-Org-Id: $remote_user
+    basicAuthSecretName: ""
+
   image:
     repository: nginx
     tag: 1.21
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [ENHANCEMENT] Define namespace in templates #184
 * [ENHANCEMENT] Use FQDN for memcached addresses #175
 * [ENHANCEMENT] Optionally generate endpoints for `X-Scope-OrgID` injection (multi-tenancy) #180
+* [ENHANCEMENT] Optionally configure Basic Auth within Nginx #205
 * [BUGFIX] Fix whitespace in runtime-config annotations, introduced in #209, fixed in #212
 * [BUGFIX] Correcting nginx config for auth orgs to right proxy_pass #192
 
--- a/templates/nginx/nginx-dep.yaml
+++ b/templates/nginx/nginx-dep.yaml
@@ -59,6 +59,11 @@ spec:
             {{- end }}
             - name: config
               mountPath: /etc/nginx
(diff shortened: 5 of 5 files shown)


ShuzZzle:
Hi @logston,

Thank you for contribution. You will also have to update the README.md with your updated values.yml entry. See the lint-test Job for more details :)

ShuzZzle (changes requested):
Please update the Readme.md with your new values.yml config option

logston:
Thanks @ShuzZzle! I've updated the PR.

README.md
```diff
@@ -590,6 +590,7 @@ Kubernetes: `^1.19.0-0`
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;client_max_body_size | string | `"1M"` |  |
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;dnsResolver | string | `"kube-dns.kube-system.svc.cluster.local"` |  |
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;setHeaders | object | `{}` |  |
+| nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;basicAuthMountPath | string | `""` | Path of `auth_basic_user_file` |
```
ShuzZzle, line 593:
```suggestion
| nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;basicAuthMountPath | string | `""` | (optional) Mount path of basic auth secret. In order to use this path, a secret with htpasswd contexts should be mounted to the container. For example:  extraVolumes:  - name: nginx-htpasswd    secret:      defaultMode: 420      secretName: nginx-htpasswd  extraVolumeMounts:  - mountPath: /keys/    name: nginx-htpasswd |
```

ShuzZzle: should be like this (must match the generated spec from values.yaml, using helm-docs)

ShuzZzle (changes requested):

Eraac:
Please also disable basic_auth for /healthz path (for kubernetes health check)

templates/nginx/nginx-config.yaml
```diff
@@ -40,6 +40,11 @@ data:
...
         {{- end }}
 
+        {{ if .Values.nginx.config.basicAuthMountPath -}}
+        auth_basic "Restricted Content";
+        auth_basic_user_file {{ .Values.nginx.config.basicAuthMountPath }};
+        {{- end }}
+
         location = /healthz {
           return 200 'alive';
         }
```
ShuzZzle, line 50:
```suggestion
        location = /healthz {
          auth_basic "off";
          return 200 'alive';
        }
```
Needs to be tested, but I think that should disable basic_auth for healthz checkpoint

ShuzZzle: @Eraac Something like this?

Eraac: @ShuzZzle I'm not 100% sure about the quote but otherwise is the only place where it is needed

README.md
logston: Oh! Many thanks.

templates/nginx/nginx-config.yaml
logston: I don't think this will have a difference given Nginx's use of [access phases](http://nginx.org/en/docs/dev/development_guide.html#http_phases). 

See this SO post: https://stackoverflow.com/questions/67975464/why-doesnt-basic-auth-work-with-a-simple-nginx-return-statement

However, I can add comment for clarity on why nginx will still return 200 in the event basic auth is enabled.

templates/nginx/nginx-config.yaml
```diff
@@ -40,7 +40,17 @@ data:
...
+        auth_basic "Restricted Content";
+        auth_basic_user_file {{ .Values.nginx.config.basicAuthMountPath }};
+        {{- end }}
+
         location = /healthz {
+          # auth_basic off is not set here as Nginx's NGX_HTTP_REWRITE_PHASE
+          # (point when this return statement is evaluated) comes before the
+          # NGX_HTTP_ACCESS_PHASE (point when basic auth is evaluated). Thus,
+          # this return statement returns a response before basic auth is
+          # evaluated.
```
ShuzZzle, line 53: I assume you tested this on your local setup? If yes, then its fine with me. Unfortunately the CI doesn't test this change

ShuzZzle (approved):

values.yaml
```diff
@@ -1165,6 +1165,19 @@ nginx:
...
+    # mounted to the container. For example:
+    #  extraVolumes:
+    #  - name: nginx-htpasswd
+    #    secret:
+    #      defaultMode: 420
+    #      secretName: nginx-htpasswd
+    #  extraVolumeMounts:
+    #  - mountPath: /keys/
+    #    name: nginx-htpasswd
+    basicAuthMountPath: ""
```
kd7lxl, line 1179: The usage here is a little awkward. If this is going to be a feature of the chart, it could do the mount work for you too. This parameter would change to something like `basicAuthSecretName: ""` and the template would get a new conditional statement in the volume mounts section.

templates/nginx/nginx-config.yaml
```diff
@@ -40,7 +40,17 @@ data:
         proxy_set_header {{ $key }} {{ $value }};
         {{- end }}
 
+        {{ if .Values.nginx.config.basicAuthMountPath -}}
+        auth_basic "Restricted Content";
+        auth_basic_user_file {{ .Values.nginx.config.basicAuthMountPath }};
```
kd7lxl, line 45: Help me understand the use case for this -- it seems pretty narrow.  It looks like this will require basic auth, but not actually use the username.

To protect the `/api/v1/push` endpoint from privilege escalation / users masquerading as other users, I would expect nginx to set the `X-Scope-OrgId` header to the basic username it received, or something like that.

Since this doesn't do that, it looks like it's only useful for creating admin users, who are expected to be trusted with all endpoints and orgs. Is that right? Is that the only use case for this?

templates/nginx/nginx-config.yaml
logston: I have. I tested it in a staging env in GCP.

values.yaml
logston: Sure thing. I can add that.

templates/nginx/nginx-config.yaml
logston: The basic auth file contains both the user name and the password in a hashed format. A series of them to be precise. Please see this link for an example: https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/

As I understanding it, `X-Scope-OrgId` is used for tenant identification, not authentication/authorization. See here: 
https://cortexmetrics.io/docs/guides/auth/

> All Cortex components take the tenant ID from a header X-Scope-OrgID on each request. A tenant (also called “user” or “org”) is the owner of a set of series written to and queried from Cortex. All Cortex components **trust this value completely: if you need to protect your Cortex installation from accidental or malicious calls then you must add an additional layer of protection.**

My particular use case requires that Prometheus installations write over the public internet to a Cortex installation. In order to provide some security (beyond TLS), I need one of the solutions that Promtheus supports in its `remote_write` configuration. The simplest is basic auth. Please see this link for a complete list: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write

kd7lxl: > As I understanding it, `X-Scope-OrgId` is used for tenant identification, not authentication/authorization.

Authentication is the process of verifying the identity of a user or process. This is why I was asking how http basic auth might tie to tenant identification.

Here (my employer), we use http basic auth to authenticate a Prometheus instance _and_ map it to the corresponding tenant with `X-Scope-OrgId`. In this way, we are doing both the authentication and authorization with http basic auth. There's no way for our http basic auth users to access each other's data.

With this change, I just want to make sure the use case is clearly documented, so that users understand that it does not relate to [Cortex auth](https://cortexmetrics.io/docs/guides/auth/) and allows http basic auth users to access each other's data. All users must be completely trusted.

kd7lxl: Maybe the way this detail could be documented is by giving an example of how one might add Cortex auth to the config, something like this:
> When using http basic auth and Cortex multi-tenancy, the username can be used as the tenant id:
> ```yaml
> setHeaders:
>   X-Scope-OrgId: $remote_user
> ```
Not sure if this will work with the current ordering of the nginx config (`setHeaders` is before auth)

logston: >  I just want to make sure the use case is clearly documented, so that users understand that it does not relate to Cortex auth

Got it. I can add that in.

logston:
Hey @kd7lxl, if the docs updates are sufficient, can I get a 👍 to merge?

values.yaml
```diff
@@ -1165,6 +1165,27 @@ nginx:
...
+    #   stringData:
+    #     .htpasswd: |
+    #       user1:$apr1$/woC1jnP$KAh0SsVn5qeSMjTtn0E9Q0
+    #       user2:$apr1$QdR8fNLT$vbCEEzDj7LyqCMyNpSoBh/
+    #
+    # Please note that the use of basic auth will not identify organizations
+    # the way X-Scope-OrgID does. Thus, the use of basic auth alone will not
+    # prevent one tenant from viewing the metrics of another. To ensure tenants
+    # are scoped appropriately, explicitly pass the `X-Scope-OrgID` header
+    # along with a basic auth header.
```
kd7lxl, line 1186: This is worded backwards... _passing_ the header requires trusting the client. Setting the header in the nginx config with appropriately scope the tenants.
```suggestion
    # prevent one tenant from viewing the metrics of another. To ensure tenants
    # are scoped appropriately, explicitly set the `X-Scope-OrgID` header
    # in the nginx config. Example
    #   setHeaders:
    #     X-Scope-Org-Id: $remote_user
```

logston: Ok. Updated.

kd7lxl (approved):

ShuzZzle (approved):

logston:
Thanks for the reviews! Unfortunately, I don't access to merge this PR:

> Only those with write access to this repository can merge pull requests.

Can either one of you merge this?

logston:
Thanks @ShuzZzle! Lastly, I see that releases are cut every 1 - 3 months or so. Is there any projected date for the next release? Many thanks.

ShuzZzle:
> Thanks @ShuzZzle! Lastly, I see that releases are cut every 1 - 3 months or so. Is there any projected date for the next release? Many thanks.

Uhm unfortunately there is no such release plan as of now. Soon™

PR #208, opened 2021-08-30
introduce cortex-helm-chart wiki

ShuzZzle:
**What this PR does**:
Introduces a new wiki specific to the cortex-helm-chart where we will document best practices and general guides 

Current test-site: https://shuzzzle.github.io/cortex-helm-chart/

**Checklist**


docs/Gemfile
```diff
@@ -0,0 +1,31 @@
+source "https://rubygems.org"
```
kd7lxl, line 1: Referencing this example: https://github.com/pmarsceill/jtd-remote, they don't include any `Gemfile` or `*.html`. My understanding is that Github Pages handles all of this for you.

ShuzZzle: I Agree but we installed a custom gem "just-the-docs" therefore my thinking was that it **might** be necessary. However that doesn't seem to be case, I will remove it then. The Layout html and the 404 html are necessary tho if we ever want custom 404 messages and/or custom layouts

ShuzZzle: Too be honest I also left it because it makes local development way easier....

docs/assets/css/styles.scss
```diff
@@ -0,0 +1,32 @@
+---
+# this ensures Jekyll reads the file to be transformed into CSS later
+# only Main files contain this front matter, not partials.
+---
+
+.code-header {
+  display: flex;
+  justify-content: flex-end;
```
kd7lxl, line 8: I don't know how flex works because I haven't done frontend for years, but this removes the extra spacing:
```suggestion
  float: right;
```

ShuzZzle: Solved!

kd7lxl (approved):

docs/_includes/codeHeader.html
```diff
@@ -0,0 +1,3 @@
+<div class="code-header">
+    <button class="copy-code-button" aria-label="Copy code to clipboard"></button>
+</div>
```
ShuzZzle, line 3: @ShuzZzle remove this file due to the div beeing auto-injected

PR #213, opened 2021-09-01
Support customizing the nginx config with values

kd7lxl:
**What this PR does**:
There are a lot of use cases for nginx and not all of them are supported by the chart. This change allows adding custom configuration to nginx, so that users can implement what they need with nginx without a change to the chart.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [FEATURE] Add autoscaler for queriers #190
 * [FEATURE] Add autoscaler for distributors #189
 * [FEATURE] Add autoscaler for ingesters #182
+* [ENHANCEMENT] Support customizing the nginx config with values #213
 * [ENHANCEMENT] Upgrade to Cortex v1.10.0 #204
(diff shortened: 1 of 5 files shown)


ShuzZzle (changes requested):
Did the CHANGELOG entry get lost while rebasing?

ShuzZzle (approved):

ShuzZzle:
yeah "let CI build the README" uhm it only triggers on changes to values.yml :(

kd7lxl:
> yeah "let CI build the README" uhm it only triggers on changes to values.yml :(

Hmm, as a PR check, we really want it to run on the full diff of the PR.

ShuzZzle (approved):

ShuzZzle:
@kd7lxl Ah great, did we just bugged it? Seems to be stuck...maybe due to the skip ci tag.

kd7lxl:
> @kd7lxl Ah great, did we just bugged it? Seems to be stuck...maybe due to the skip ci tag.

Fixed in #217

PR #216, opened 2021-09-07
Bug/support tls service monitors

### LINKED ISSUE #215: Chart is missing support for pulling metrics with TLS
I enabled TLS in Cortex, the service monitors also need the ability to pull metrics over TLS.

danfromtitan:
**What this PR does**:
Add TLS configuration to ServiceMonitors templates.

Fixes #215

**Checklist**


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * [ENHANCEMENT] Use FQDN for memcached addresses #175
 * [ENHANCEMENT] Optionally generate endpoints for `X-Scope-OrgID` injection (multi-tenancy) #180
 * [ENHANCEMENT] Optionally configure Basic Auth within Nginx #205
+* [ENHANCEMENT] Add service monitors TLS support #215
```
ShuzZzle, line 16:
```suggestion
* [ENHANCEMENT] Add service monitors TLS support #216
```

wrong PR number

ShuzZzle:
Hi @danfromtitan,

Just a heads up. We have not forgot about your PR but considering a more dynamic approach to extending the ServiceMonitor options since there are so so many different options.

Maybe consider having something like this
```yaml
{{- with .Values.ingester.serviceMonitor.extraEndpointSpec }}
{{ tpl . $ | nindent <LEVEL> }}
{{- end }}
```
```yaml
# values.yml
    extraEndpointSpec:
      scheme: HTTPS
      tlsConfig: foo
```

maybe even for the whole [ServiceMonitorSpec](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#servicemonitorspec) since there are so many different options

danfromtitan:
I can see the value of adding a generic extraEndpointSpec, but taking it at the ServiceMonitorSpec level it too much IMO. Not only the configuration becomes hard to figure out for a newcomer, but you also have different selector.matchLabels for each component that you'd want to feed from the template.

danfromtitan:
@ShuzZzle, I added extraEndpointSpec as discussed above.

kd7lxl (approved):

ShuzZzle (approved):

PR #224, opened 2021-09-21
Update memcached chart

iamashwanth:
**What this PR does**:

Newer version of the chart allows configuration of liveness and
readiness probe on metrics container.

Defaults provided by the chart might not be ideal, using the newer
version allows the end user to tweak the probe behaviour.

 -

**Checklist**
- [X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`


ShuzZzle:
Thanks but we already have renovate for that

iamashwanth:
@ShuzZzle Didn't know about this. Thanks!

PR #233, opened 2021-10-04
update container names to match the ones expected by cortex-mixin

danfromtitan:
**What this PR does**:
All Cortex container names take the Chart.Name value, making it impossible to query metrics based on _container_ metric label. On top of that, the [cortex-mixin configuration](https://github.com/grafana/cortex-jsonnet/blob/main/cortex-mixin/config.libsonnet#L28) looks for specific names. The change aligns the container names with their functionality.

Fixes #233 

**Checklist**
- [X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`


templates/nginx/nginx-dep.yaml
```diff
@@ -44,7 +44,7 @@ spec:
       {{- end}}
       {{- end }}
       containers:
-        - name: nginx
+        - name: cortex-gw
```
ShuzZzle, line 47: Our nginx is not really the same as the cortex-mixin "cortex-gw". Grafana has its own solution, we also lack the metrics. This won't really work so might as well keep the name "nginx"

ShuzZzle (changes requested):

templates/nginx/nginx-dep.yaml
danfromtitan: rolledback nginx container name

ShuzZzle (approved):

ShuzZzle:
LGTM

templates/alertmanager/alertmanager-dep.yaml
```diff
@@ -98,7 +98,7 @@ spec:
             - name: storage
               mountPath: "/data"
         {{- end}}
-        - name: {{ .Chart.Name }}
+        - name: alertmanager
```
kd7lxl, line 101: container `name: {{ .Chart.Name }}` is a [typical way](https://github.com/helm/helm/blob/d4cc130fa90027e8c1c69ad2a06c6dee4c31d29d/pkg/chartutil/create.go#L308) to name containers in helm.

The cortex-mixin looks like it's matching `job_name`, not `container`, so it's unclear to me if this PR solves your problem.

The pods already have the standard label to identify the component (e.g., `app.kubernetes.io/component=alertmanager`). Can you use that, maybe with a relabel config, to do what you need?

ShuzZzle:
Hey @danfromtitan,

I use the mixin as well. I can share my config in order to get it working. I use some relabel config to the cluster part working and slighty adjusted jobnames. I did not have to rename containers, that really doesn't make super much sense too me

templates/alertmanager/alertmanager-dep.yaml
danfromtitan: There are lots of places in cortex-mixin such [as this](https://github.com/grafana/cortex-jsonnet/blob/main/cortex-mixin/dashboards/dashboard-utils.libsonnet#L482) where container name label is part of a promql. Most of the time these variables resolve from hardcoded strings ('compactor' [in the example above](https://github.com/grafana/cortex-jsonnet/blob/main/cortex-mixin/dashboards/compactor-resources.libsonnet#L31)) which won't render the right result unless the container comes with the specific name.

In other cases even when the query returns successfully, the result is incorrect because it returns all occurrences in the namespace with a common container name matching Chart.Name, so you'd want to be specific to ensure you look at the right results for the component.

I do agree this change puts the cart before the horse, in other words the dashboards should be properly implemented instead of changing the Helm chart to support faulty dashboards, but the change was small and I didn’t think it would break anything. 

Sure, if I would implement dashboards from scratch I would use job_name and not care about container. My goal was to re-use cortex-mixin's with the minimal amount of changes required to load all dashboards succesfully while deploying from the Helm chart. Container name was the only one I couldn't workaround without a change in the chart.

danfromtitan:
>  I can share my config in order to get it working

Thanks for sharing, I was [looking for this relabeling detail](https://github.com/grafana/cortex-jsonnet/issues/361#issuecomment-902752745) earlier, I think it would reduce the amount of changes required in cortex-mixin fork but it would still need container to have the component in the name for the reasons mentioned above.

ShuzZzle:
> >  I can share my config in order to get it working
> 
> Thanks for sharing, I was [looking for this relabeling detail](https://github.com/grafana/cortex-jsonnet/issues/361#issuecomment-902752745) earlier, I think it would reduce the amount of changes required in cortex-mixin fork but it would still need container to have the component in the name for the reasons mentioned above.

Yep I agree. I also didn't check every dashboard, so there still might be some problems. I feel like this is a bit out of scope but maybe we can figure something out. Thanks for raising the Issue

danfromtitan:
I did, the relabelling helps whenever a namespace/job value comes into play, but container labels are hard coded as I [mentioned above](https://github.com/cortexproject/cortex-helm-chart/pull/233#discussion_r721757208).

Besides the relabelling and the container names change in the chart, you can [see here](https://github.com/danfromtitan/cortex-jsonnet/compare/main...danfromtitan:helm-supported-dashboards) the additional changes required to get all dashboards fully functional.

ShuzZzle:
> I did, the relabelling helps whenever a namespace/job value comes into play, but container labels are hard coded as I [mentioned above](https://github.com/cortexproject/cortex-helm-chart/pull/233#discussion_r721757208).
> 
> Besides the relabelling and the container names change in the chart, you can [see here](https://github.com/danfromtitan/cortex-jsonnet/compare/main...danfromtitan:helm-supported-dashboards) the additional changes required to get all dashboards fully functional.

Have you created a PR over at cortex-jsonnet to be more dynamic with the mixin?

danfromtitan:
> Have you created a PR over at cortex-jsonnet to be more dynamic with the mixin?

That was my next target, but I cannot test the changes there unless this PR is merged first.

ShuzZzle:
> > Have you created a PR over at cortex-jsonnet to be more dynamic with the mixin?
> 
> That was my next target, but I cannot test the changes there unless this PR is merged first.

Excuse my ignorance but why? Which dashboards require a change here? I don't really get it

danfromtitan:
> Excuse my ignorance but why? Which dashboards require a change here? I don't really get it

Please see in my [comment above](https://github.com/cortexproject/cortex-helm-chart/pull/233#discussion_r721757208).

ShuzZzle:
> > Excuse my ignorance but why? Which dashboards require a change here? I don't really get it
> 
> Please see in my [comment above](https://github.com/cortexproject/cortex-helm-chart/pull/233#discussion_r721757208).

Gotcha! I will talk to @kd7lxl

kd7lxl:
I'm a little puzzled by the conversation here.

> I cannot test the changes there unless this PR is merged first.

This doesn't make any sense to me. You can `helm install` from your branch just as easily as you can `helm install` from the master branch or a release.

Please test the desired dashboard behavior against this branch to prove that it solves the problem. Please share your results.

If everything works as expected, we can target this for v1.0.0. It will need to be in a major version release because it's a breaking change -- everyone who has dashboards targeting the current container names will need to rewrite their dashboards.

danfromtitan:
>  everyone who has dashboards targeting the current container names will need to rewrite their dashboards

In case someone's using [cortex-mixin](https://github.com/grafana/cortex-jsonnet/tree/main/cortex-mixin) to create dashboards, they will need the same change I requested in this PR.

Anyone else using container label in their dashboards would look at misleading results, because container label "cortex" would query the metrics from all components instead of the specific component dashboard is intended for.

I have made [changes to the cortex-mixin project](https://github.com/danfromtitan/cortex-jsonnet/compare/main...danfromtitan:helm-supported-dashboards) that are waiting to be merged, dahboards work perfectly, we need this PR merged first before I can submit the other.

ShuzZzle:
@danfromtitan have you opened up a PR with the folks over at grafana?

danfromtitan:
Not yet, it's on my list. Thanks for merging this.

danfromtitan:
> @danfromtitan have you opened up a PR with the folks over at grafana?

Thanks for accepting this PR, I opened https://github.com/grafana/cortex-jsonnet/pull/427 to fix the outstanding issues with Grafana dashboards when Cortex is deployed from the Helm chart.

PR #236, opened 2021-10-05
add option to set ingesters podManagementPolicy

AndreZiviani:
**What this PR does**:
Add option to specify `podManagementPolicy` when using ingesters as statefulset

**Checklist**


AndreZiviani:
I believe the recommended option is to have it as `Parallel`, should I change the default option?

Cortex mixin have it as `Parallel` by default
https://github.com/grafana/cortex-jsonnet/blob/c1244964dbbc7901a8ef13327c157d6b1fac7e8d/cortex/tsdb.libsonnet#L126

More info: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies
```
Parallel pod management tells the StatefulSet controller to launch or terminate all Pods in parallel, and to not wait for Pods to become Running and Ready or completely terminated prior to launching or terminating another Pod. This option only affects the behavior for scaling operations. Updates are not affected.
```

ShuzZzle:
> I believe the recommended option is to have it as `Parallel`, should I change the default option?
> 
> Cortex mixin have it as `Parallel` by default https://github.com/grafana/cortex-jsonnet/blob/c1244964dbbc7901a8ef13327c157d6b1fac7e8d/cortex/tsdb.libsonnet#L126
> 
> More info: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies `Parallel pod management tells the StatefulSet controller to launch or terminate all Pods in parallel, and to not wait for Pods to become Running and Ready or completely terminated prior to launching or terminating another Pod. This option only affects the behavior for scaling operations. Updates are not affected.`

I guess we could do that, will probably speed up the scaling significantly. @kd7lxl  Thoughts?

kd7lxl:
> @kd7lxl Thoughts?

From the docs https://cortexmetrics.io/docs/guides/ingesters-scaling-up-and-down/#scaling-down:
> * If two or more ingesters are scaled down in a short period of time, queriers may miss some data at query time due to series that were stored in the terminated ingesters but their blocks haven’t been discovered yet.
> * Ingesters should be scaled down one by one

So scale-down can't be parallel.

Do you have a strong use case for `Parallel`? Given it breaks the read path, I can't see it being useful in production.

If we were going to accept this, it would need a big warning so that people don't accidentally use it outside of very specific testing environments.

ShuzZzle:
> > @kd7lxl Thoughts?
> 
> From the docs https://cortexmetrics.io/docs/guides/ingesters-scaling-up-and-down/#scaling-down:
> 
> > * If two or more ingesters are scaled down in a short period of time, queriers may miss some data at query time due to series that were stored in the terminated ingesters but their blocks haven’t been discovered yet.
> > * Ingesters should be scaled down one by one
> 
> So scale-down can't be parallel.
> 
> Do you have a strong use case for `Parallel`? Given it breaks the read path, I can't see it being useful in production.
> 
> If we were going to accept this, it would need a big warning so that people don't accidentally use it outside of very specific testing environments.

fine with me.

AndreZiviani:
> Do you have a strong use case for Parallel? Given it breaks the read path, I can't see it being useful in production.
Grafana is using this for more than a year
https://github.com/grafana/cortex-jsonnet/pull/143

Scaling down the ingesters is already a sensitive operation that needs to be done manually (AFAIK you can't just decrease the number os replicas, as stated on the link you sent)

Anyway, thats why I left the default value as is, but I would like the option to change it

Loki also have it enabled by default:
Mixin: https://github.com/grafana/loki/blob/main/production/ksonnet/loki/ingester.libsonnet#L62
Helm: https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/templates/ingester/statefulset-ingester.yaml#L10
PR: https://github.com/grafana/helm-charts/pull/448

kd7lxl:
> Scaling down the ingesters is already a sensitive operation that needs to be done manually (AFAIK you can't just decrease the number os replicas, as stated on the link you sent)

This chart does support scale-down and implements the strategy required by the docs:
https://github.com/cortexproject/cortex-helm-chart/blob/7b913471487ecbe32f5825a00b92ac76785e58b4/values.yaml#L566-L572

Please document the `podManagementPolicy` value with the `# --` style to include an appropriate warning.

AndreZiviani:
done

PR #255, opened 2021-11-12
pass distributer service type from values.yaml to support other service types like LoadBalancer

### LINKED ISSUE #254: pass distributer service type from values.yaml to support other service types like LoadBalancer
As of now distributor service type is hardcoded as  ClusterIP, so in order to make  cortex-distributer available outside the cluster e:g for other accounts or clusters we have to edit the already created service to make it type of `LoadBalancer`.

Creating this issue so we can share PR to address this and distributer service type can be overriden.

ahsan-n:
**What this PR does**:
enable users to override distributer service type
Fixes #254

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 * [BUGFIX] Fix nil pointer evaluation when using `ruler.dictonaries` option #242
 * [DEPENDENCY] Update Helm release memcached to v5.15.5 #241
 * [DEPENDENCY] Update Helm release memcached to v5.15.8 #247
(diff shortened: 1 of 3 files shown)


ShuzZzle:
Sorry but we can kinda already have this behaviour by using nginx. Usually its like this:

- Install ingress controller to your liking with service type loadbalancer
- create ingress resource and point to nginx server
- voila

I do not really see a need to talk to the distributor directly

ahsan-n:
Thanks for review i will look at it. 🙂

PR #258, opened 2021-11-17
- Fixed hardcoding of containerSecurityContext in Ingester statefulset

RyanW8:
- Brought setting of containerSecurityContext for the Ingester statefulset inline with other components configuration

**What this PR does**:

- Removes the hardcoded value for containerSecurityContext in the ingester statefulset
- Brough the configuration of the ingester containerSecurityContext inline with what is documented in the values.

Didn't raise an issue

### DIFF
--- a/templates/ingester/ingester-statefulset.yaml
+++ b/templates/ingester/ingester-statefulset.yaml
@@ -150,8 +150,6 @@ spec:
             {{- toYaml .Values.ingester.readinessProbe | nindent 12 }}
           resources:
             {{- toYaml .Values.ingester.resources | nindent 12 }}
-          securityContext:
-            readOnlyRootFilesystem: true
           env:
             {{- if .Values.ingester.env }}
               {{- toYaml .Values.ingester.env | nindent 12 }}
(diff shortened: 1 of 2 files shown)


templates/ingester/ingester-statefulset.yaml
```diff
@@ -151,8 +150,6 @@ spec:
             {{- toYaml .Values.ingester.readinessProbe | nindent 12 }}
           resources:
             {{- toYaml .Values.ingester.resources | nindent 12 }}
-          securityContext:
-            readOnlyRootFilesystem: true
```
ShuzZzle, line 155: This should be
```yaml
          {{- if .Values.ingester.containerSecurityContext.enabled }}
          securityContext: {{- omit .Values.ingester.containerSecurityContext "enabled" | toYaml | nindent 12 }}
          {{- end }}
```

ShuzZzle (changes requested):

templates/ingester/ingester-statefulset.yaml
kd7lxl: Thanks for catching this. We want to retain the default of `readOnlyRootFilesystem: true` for sure.

ShuzZzle (approved):

kd7lxl (approved):

PR #264, opened 2021-11-25
Optionally set service account per dep/statefulset.

### LINKED ISSUE #251: Support separate service accounts for each cortex service
I'm looking into using [gke workload identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) to assign gcp iam permissions to k8s pods, and I would like to be able to configure different service accounts per cortex service. For example, nginx doesn't need access to gcs, so ideally I'd use a different service account for nginx than for the ingesters.

Would you be open to a patch allowing users to override service account names by service? Happy to submit something if so.

jmcarp:
**What this PR does**:

Different cortex components might require different cloud provider
permissions. When using gke workload identity, pod iam permissions are
managed by kubernetes service account. This patch allows operators to
use different service accounts for each deployment or statefulset to
support fine-grained control of iam permissions in this scenario.

Fixes #251

**Checklist**

### DIFF
--- a/templates/alertmanager/alertmanager-dep.yaml
+++ b/templates/alertmanager/alertmanager-dep.yaml
@@ -35,7 +35,7 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
     spec:
-      serviceAccountName: {{ template "cortex.serviceAccountName" . }}
+      serviceAccountName: {{ .Values.alertmanager.serviceAccount.name | default (include "cortex.serviceAccountName" . ) }}
     {{- if .Values.alertmanager.priorityClassName }}
       priorityClassName: {{ .Values.alertmanager.priorityClassName }}
     {{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -177,6 +177,10 @@ alertmanager:
     annotations: {}
     labels: {}
 
+  serviceAccount:
+    # -- "" disables the individual serviceAccount and uses the global serviceAccount for that component
+    name: ""
+
   serviceMonitor:
     enabled: false
     additionalLabels: {}
@@ -317,6 +321,10 @@ distributor:
     annotations: {}
     labels: {}
 
+  serviceAccount:
+    # -- "" disables the individual serviceAccount and uses the global serviceAccount for that component
+    name: ""
+
   serviceMonitor:
     enabled: false
     additionalLabels: {}
@@ -421,6 +429,9 @@ ingester:
   service:
     annotations: {}
     labels: {}
+
+  serviceAccount:
+    name:
 
   serviceMonitor:
     enabled: false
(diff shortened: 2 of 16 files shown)


values.yaml
```diff
@@ -211,6 +211,9 @@ alertmanager:
     annotations: {}
     labels: {}
 
+  serviceAccount:
+    name:
```
ShuzZzle, line 215:
```suggestion
    # -- "" disables the individual serviceAccount and uses the global serviceAccount for that component
    name: ""
```

Small suggestion, I would prefer that than having no value at all. Looks odd to me

ShuzZzle:
Oh and please rebase your PR's

ShuzZzle (approved):

templates/alertmanager/alertmanager-dep.yaml
```diff
@@ -35,7 +35,7 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
     spec:
-      serviceAccountName: {{ template "cortex.serviceAccountName" . }}
+      serviceAccountName: {{ with .Values.alertmanager.serviceAccount.name }}{{ . }}{{ else }}{{ template "cortex.serviceAccountName" . }}{{ end }}
```
kd7lxl, line 38: Is there any specific benefit to using `with`/`else` here rather than [`| default`](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function)?

ShuzZzle: ah nice idea! but it should be
```yaml
{{ .Values.alertmanager.serviceAccount.name | default (include "cortex.serviceAccountName" .) }}
```
right?

templates/alertmanager/alertmanager-dep.yaml
```diff
@@ -35,7 +35,7 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
     spec:
-      serviceAccountName: {{ template "cortex.serviceAccountName" . }}
+      serviceAccountName: {{ .Values.alertmanager.serviceAccount.name | default (template "cortex.serviceAccountName" . ) }}
```
ShuzZzle, line 38:
```suggestion
      serviceAccountName: {{ .Values.alertmanager.serviceAccount.name | default (include "cortex.serviceAccountName" . ) }}
```

ShuzZzle (changes requested):
should be include in order for it to work

templates/alertmanager/alertmanager-dep.yaml
jmcarp: Thanks, applied this throughout.

kd7lxl (approved):

ShuzZzle (approved):

PR #275, opened 2021-11-29
Enable bucket index by default.

jmcarp:
**What this PR does**:

The cortex docs recommend enabling the bucket index for both the querier
and the store-gateway: https://cortexmetrics.io/docs/blocks-storage/production-tips/.

Any reason not to enable this by default?

Fixes #<issue number>

**Checklist**


ShuzZzle (approved):
nope no specific reason as to why it's not enabled

kd7lxl (approved):

PR #280, opened 2021-12-01
Optionally pass cortex config as configmap.

### LINKED ISSUE #235: Deploy the config as a `configMap` rather than a secret
Let's also add the ability to deploy the config as a `configMap` rather than a secret?

_Originally posted by @cabrinha in https://github.com/cortexproject/cortex-helm-chart/issues/227#issuecomment-930516408_

And have the credentials passed as environment variables backed by secrets? That would be quite optimal.

jmcarp:
**What this PR does**:
Optionally pass cortex config as configmap.

Note: I only updated the ingestor templates so far, to verify that the approach makes sense. If it does, I'll copy and paste the updated volume to the other templates.

Fixes #235.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## master / unreleased
+* [FEATURE] Optionally manage cortex config as configmap. #280
 
 ## 1.1.0 / 2021-12-01
 
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -1,0 +1,12 @@
+{{- if (and (not .Values.useExternalConfig) (.Values.useConfigMap)) }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "cortex.fullname" . }}-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "cortex.labels" . | nindent 4 }}
+data:
+  cortex.yaml: |
+    {{- tpl (toYaml .Values.config) . | nindent 4 }}
+{{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -34,6 +34,7 @@ serviceAccount:
   annotations: {}
   automountServiceAccountToken: true
 
+useConfigMap: false
 useExternalConfig: false
 externalConfigSecretName: 'secret-with-config.yaml'
 externalConfigVersion: '0'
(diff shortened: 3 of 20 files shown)


ShuzZzle:
Nice!

with that approach however, we should really have the s3 credentials (or similiar) **not** in the configMap

values.yaml
```diff
@@ -34,6 +34,7 @@ serviceAccount:
   annotations: {}
   automountServiceAccountToken: true
 
+useConfigMap: true
```
ShuzZzle, line 37: I wouldn't enable this per default personally

ShuzZzle (changes requested):
Hey, thank you for your contribution.

Can you please add the CHANGELOG.md entry?
Also we would need to add a mechanism similiar to: [This](https://github.com/cortexproject/cortex-helm-chart/blob/master/templates/ingester/ingester-dep.yaml#L33) 

So we can restart components when configmap changes, right now we can't use something like configmap reloader AFAIK

jmcarp:
Updated, lmk if I missed anything.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * [CHANGE] Disable ingester liveness probes by default. #263
 * [BUGFIX] workaround for .Capabilities.APIVersions. #277
 * [DEPENDENCY] Update Helm release memcached to v5.15.9 #273
+* [FEATURE] Optionally manage cortex config as configmap. #280
```
ShuzZzle, line 12: technically FEATURE's should be the first entry in the changelog

ShuzZzle (commented):
I will have another glance tomorrow but this looks really good. Thanks!

ShuzZzle (approved):
I re-edited the guide, because inline code blocks completely broke the the text unfortunately. We will have to adjust the markdown processor something like that in the future

templates/configmap.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if (and (not .Values.useExternalConfig) (.Values.useConfigMap)) }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "cortex.fullname" . }}
```
ShuzZzle, line 5:
```suggestion
  name: {{ template "cortex.fullname" . }}-config
```
tiny improvement idea

ShuzZzle:
@jmcarp Please be careful when rebasing, my changes to the guide were lost

kd7lxl (approved):

jmcarp:
Ack, sorry, not used to collaborating on a branch 😓

PR #288, opened 2021-12-13
Use port numbers instead of names in prometheus annotations.

jmcarp:
Prometheus scrape configs often use the `prometheus.io/port` annotation
to configure targets per pod:

```
- source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
  action: replace
  regex: ([^:]+)(?::\d+)?;(\d+)
  replacement: $1:$2
  target_label: __address__
```

This rule only matches when the `prometheus.io/port` annotation is a
port number, and it's not clear to me how the scrape config could be
modified to work with port names. To make the chart compatible with
standard prometheus scrape configs, this patch changes the default port
annotation from port names to port numbers.

The specific issue that I noticed here was that prometheus was scraping the memberlist port on all pods and polluting the logs with many messages like `"unknown message type" msgType=G`. It sounds like using `serviceMonitor` would resolve this, but I don't use the prometheus operator at the moment.

Discussed at https://cloud-native.slack.com/archives/CCYDASBLP/p1639277366459700.

Signed-off-by: Josh Carp <jm.carp@gmail.com>

**What this PR does**:

Fixes #<issue number>

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## master / unreleased
+* [CHANGE] Use port number for prometheus port annotations. #288
 * [CHANGE] Disable ingester startup probes by default. #286
 * [FEATURE] Optionally manage cortex config as configmap. #280
 * [ENHANCEMENT] Upgrade to Cortex v1.11.0 #272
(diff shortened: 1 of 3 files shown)


ShuzZzle (approved):
LGTM. Didn't my example help?

Anyway... Could you also remove (LINE 1108, nginx podAnnotations)
```yaml
  podAnnotations:
    prometheus.io/scrape: ''
    prometheus.io/port: 'http-metrics'
```

should be just
```yaml
podAnnotations: {}
```

nginx has no metrics anyway so we should remove the annotation

ShuzZzle:
> Basically, I want the chart to do the right thing and not scrape random ports when used with vanilla prometheus and the standard scrape configs. And using named ports in `prometheus.io/port` confused me, since I thought prometheus would be able to use those annotations, which apparently it can't.

Yeah that part confused me as well. But looking at relabel_configs it doesn't really matter how the label is called it could also be
`prometheus.evil/port: 900` as long as the relabel step for the `__address__` makes sense and I agree it's just easier with numbers instead of port names

kd7lxl (approved):

PR #297, opened 2022-01-04
rework nginx config

nschad:
**What this PR does**:

- 100% Implementation of Cortex v1 and legacy API (not done)
- Trying to implement `/` `/config` `/runtime_config` `/services` `/ready` `/metrics` `/debug/pprof` `/debug/fgprof` for all services

Fixes #<issue number>

**Checklist**

### DIFF
--- a/templates/nginx/all-services-config.yaml
+++ b/templates/nginx/all-services-config.yaml
@@ -1,0 +1,39 @@
+{{- if .Values.nginx.enabled }}
+{{- $rootDomain := printf "%s.svc.%s:%d" .Release.Namespace .Values.clusterDomain (.Values.config.server.http_listen_port | int) }}
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: {{ include "cortex.nginxFullname" . }}-all-services-page
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "cortex.nginxLabels" . | nindent 4 }}
+data:
+  all-services.html: |-
+    <html>
+    <script>
+        function addURI(elem) {
+            var requestURI = location.pathname.substr(1)
+            var href = new URL(requestURI, elem.getAttribute("href")).href
+            var url = "http://{{ .Values.nginx.redirectUrl }}?redirect=" + href
+            elem.setAttribute("href", url)
+    }
+    </script>
+    <head>
+    <title>Cortex All Services</title>
+    <body>
+    <h1>Cortex All Services</h1>
+    <hr>
+    <ul>
+        <li><a href="http://{{ template "cortex.alertmanagerFullname" . }}.{{ $rootDomain }}" onclick="addURI(this)">Alertmanager</a></li>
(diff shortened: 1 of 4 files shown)


templates/nginx/all-services-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+    {{- include "cortex.nginxLabels" . | nindent 4 }}
+data:
+  all-services.html: |-
+    <html>
+    <script>
+        function addURI(elem) {
+            var requestURI = location.pathname.substr(1)
+            var href = new URL(requestURI, elem.getAttribute("href")).href
+            var url = "http://{{ .Values.nginx.redirectUrl }}?redirect=" + href
+            elem.setAttribute("href", url)
```
ShuzZzle, line 18: my idea is that the microservice k8 url + e.g. `/config` would be returned back to nginx so nginx can open a proxy_pass connection without having us to define 100 different endpoints.

This actually works fine in my test's but I'm unsure if that this is the right/correct way

templates/nginx/nginx-config.yaml
```diff
@@ -67,6 +69,22 @@ data:
...
+          try_files '' /all-services.html =404;
+        }
+
+        location /config {
+          if ($arg_redirect) {
+            proxy_pass $arg_redirect;
+          }
+
+          try_files '' /all-services.html =404;
+        }
```
ShuzZzle, line 86: Explanation or the idea behind that is:

First redirect to all-services page. User determines the correct service and service + `/config` is returned back where it now enters the if clause and sends to to that microservice + `/config`

values.yaml
```diff
@@ -1046,6 +1046,7 @@ nginx:
   enabled: true
   replicas: 2
   http_listen_port: 80
+  redirectUrl: ""
```
ShuzZzle, line 1049: basically the nginx FQDN without http://

templates/nginx/all-services-config.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if .Values.nginx.enabled }}
```
kd7lxl, line 1: In my env I would not want these admin pages accessible to most users. The simplest way to achieve this would be to just disable the service index. Can you make it optional?
```suggestion
{{- if and .Values.nginx.enabled .Values.nginx.serviceIndex.enabled }}
```

ShuzZzle: hmm I can absolutely do that (not fully done anyway) but I don't know really know what qualifies as an admin page. I really wanted to do everything (including supporting delete requests) and then let the user decide what is best for his corp. So either

- Use at is it
- Add Authentication & Authorization infront of it (Should be done anyway)
- Completely roll your own

kd7lxl: Yes, I would leave authorization up to the user. My requests here was to provide a mechanism to disable this feature entirely, like how the chart is today. Since it's not required to run Cortex it can be an optional feature of the chart.

ShuzZzle: I mean yeah sure. If we could maybe define what would be considered an "admin endpoint" then we could make these optional for the user?

PR #302, opened 2022-01-15
move from quay.io/kiwigrid/k8s-sidecar to omegavvweapon/kopf-k8s-side…

ShuzZzle:
Switches the current used sidecar to use a "proper" kubernetes-style operator image

**What this PR does**:
See Comment [#219 ](https://github.com/cortexproject/cortex-helm-chart/pull/219#issuecomment-1005989208)

**Checklist**

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -308,24 +308,26 @@ alertmanager:
     enableUniqueFilenames: false
     # -- Label that should be used for filtering
     label: cortex_alertmanager
-    # -- Determines how kopf-k8s-sidecar will run. If WATCH it will run like like a normal operator forever.
+    # -- Determines how kopf-k8s-sidecar will run. If WATCH it will run like a normal operator forever.
     # If LIST it will gather the matching configmaps and secrets currently present,
     # write those files to the destination directory and die
-    watchMethod: null
+    watchMethod: ""
     # -- The value for the label you want to filter your resources on.
     # Don't set a value to filter by any value
-    labelValue: null
+    labelValue: ""
     # -- Folder where the files should be placed.
     folder: /data
     # -- The default folder name, it will create a subfolder under the `folder` and put rules in there instead
-    defaultFolderName: null
+    defaultFolderName: ""
     # -- The Namespace(s) from which resources will be watched.
     # For multiple namespaces, use a comma-separated string like "default,test".
     # If not set or set to ALL, it will watch all Namespaces.
-    searchNamespace: null
+    searchNamespace: ""
     # -- The annotation the sidecar will look for in ConfigMaps and/or Secrets to override the destination folder for files.
     # If the value is a relative path, it will be relative to FOLDER
-    folderAnnotation: null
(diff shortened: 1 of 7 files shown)


ShuzZzle:
Hi @gburton1 @amouhadi,

May I request a review from either of you regarding this? Technically nothing really changed apart from some documentation but it would be cool if you can have a look.

gburton1:
I only have my phone this weekend, so pr reviews are not very easy to look at, but overall it looks good to me. If you haven’t merged by Tuesday, I’ll take a closer look.

values.yaml
```diff
@@ -655,39 +670,39 @@ ruler:
...
+    # -- Folder where the files should be placed.
     folder: /tmp/rules
     # -- The default folder name, it will create a subfolder under the `folder` and put rules in there instead
     defaultFolderName: null
-    # -- If specified, the sidecar will search for rules config-maps inside this namespace.
-    # Otherwise the namespace in which the sidecar is running will be used.
-    # It's also possible to specify ALL to search in all namespaces
+    # -- The Namespace(s) from which resources will be watched.
+    # For multiple namespaces, use a comma-separated string like "default,test".
+    # If not set or set to ALL, it will watch all Namespaces.
```
ShuzZzle, line 702:
```suggestion
    # If not set or set to ALL, it will watch all Namespaces.
```
that's different than before.

values.yaml
```diff
@@ -655,39 +670,39 @@ ruler:
...
+    # -- A value of true will produce unique filenames to avoid issues when duplicate data keys exist between ConfigMaps
+    # and/or Secrets within the same or multiple Namespaces.
     enableUniqueFilenames: false
-    enabled: false
     # -- label that the configmaps with rules are marked with
     label: cortex_rules
+    # -- Determines how kopf-k8s-sidecar will run. If WATCH it will run like like a normal operator forever.
+    # If LIST it will gather the matching configmaps and secrets currently present,
+    # write those files to the destination directory and die
     watchMethod: null
```
ShuzZzle, line 692:
```suggestion
    watchMethod: ""
```

I don't think we use `null` anywhere else. maybe we should change that for consistency

values.yaml
gburton1: Yeah, that's what I was saying in my comment from Friday. This is the only gotcha in this alternative sidecar. Strangely, the author decided to break compatibility on this one parameter 🤷 . It's the only sidecar parameter that we currently override from the current defaults here in the Cortex Helm chart. We did not want the new default behavior to watch all namespaces. So if you wanted to make this a totally backwards compatible change, then you would need to set it explicitly to `{{ .Release.Namespace }}`. Not sure it's worth that effort though. It's a pretty easy release note for an existing user to adjust to.

ShuzZzle: Yeah I don't see it as a big problem either. The guide has it as an example and its well documented

gburton1 (approved):

kd7lxl (approved):

PR #304, opened 2022-01-24
rollback optimization from release 1.0.0

ShuzZzle:
**What this PR does**:
Removes optimization from querier route logic and applies the default values. The values appear to be wrong to begin with and should have been:
```yaml
query_store_after: 6h
query_ingesters_within: 7h
```

but you actually have to calculate them to get decent values which is out of scope for this helm-chart. See [Production Tips](https://cortexmetrics.io/docs/blocks-storage/production-tips/#how-to-estimate--querierquery-store-after) on how to do that.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 * [CHANGE] move from quay.io/kiwigrid/k8s-sidecar to omegavvweapon/kopf-k8s-sidecar image #302
 * [CHANGE] undo querier store optimization #304
-  * config.querier.query_ingesters_within: 13h -> 0s
-  * config.querier.query_store_after: 12h -> 0s
+  * config.querier.query_ingesters_within: 13h -> 0s (default)
+  * config.querier.query_store_after: 12h -> 0s (default)
 
 ## 1.2.0 / 2021-12-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## master / unreleased
 
 * [CHANGE] move from quay.io/kiwigrid/k8s-sidecar to omegavvweapon/kopf-k8s-sidecar image #302
+* [CHANGE] undo querier store optimization #304
+  * config.querier.query_ingesters_within: 13h -> 0s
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -115,12 +115,11 @@ config:
...
     # -- Maximum lookback beyond which queries are not sent to ingester. 0 means all
-    # queries are sent to ingester. Ingesters by default have no data older than 12 hours,
-    # so we can safely set this 13 hours
-    query_ingesters_within: 13h
+    # queries are sent to ingester.
+    query_ingesters_within: 0s
     # -- The time after which a metric should be queried from storage and not just
     # ingesters.
-    query_store_after: 12h
+    query_store_after: 0s
```
kd7lxl, line 122: Based on https://cortexmetrics.io/docs/blocks-storage/production-tips/#avoid-querying-non-compacted-blocks, I calculated it should be more like this:
```yaml
querier:
  query_ingesters_within: 5h50m
  query_store_after: 5h45m
```
Specifically, it should be less than `block_storage.tsdb.retention_period`, so that it's querying from block storage before it is flushed from the ingesters.

ShuzZzle: yeah I don't know @kd7lxl, I think we should just do the defaults. We can't really estimate how long step X takes for each given deployment.

kd7lxl: Sure, if we want the defaults we should remove these lines rather than redefining the current defaults.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## master / unreleased
 
 * [CHANGE] move from quay.io/kiwigrid/k8s-sidecar to omegavvweapon/kopf-k8s-sidecar image #302
+* [CHANGE] undo querier store optimization #304
+  * config.querier.query_ingesters_within: 13h -> 0s
+  * config.querier.query_store_after: 12h -> 0s
```
kd7lxl, line 8: Might be helpful to people to note that these are defaults.

kd7lxl (approved):
As a follow-up, it would be useful to provide some recommended values for these like in the block storage guide.

kd7lxl (approved):

PR #310, opened 2022-01-30
update sidecar docs

nschad:
**What this PR does**:
Clarify sidecar docs

Fixes #<issue number>

**Checklist**

### DIFF
--- a/docs/guides/configure_rules_via_sidecar.markdown
+++ b/docs/guides/configure_rules_via_sidecar.markdown
@@ -13,6 +13,8 @@ This feature is disabled by default. Here is a simple example:
 Please not that this is only supported with the **local** backend. Otherwise cortex will overwrite what the operator puts in the folder with what is in s3/gcs/azure/swift. Cortex does not 2-way sync the files. Also please be aware of the following:
 - local alertmanager config storage is read-only (cortex API is severely limited to read-only requests)
 - local alertmanager storage does not support state persistency (however your configurations are saved in etcd via kubernetes)
+- local ruler storage does not support RuleGroups
+- local ruler storage does not support namespaces
 ```yaml
 backend: "local"
 ```
--- a/docs/guides/configure_rules_via_sidecar.markdown
+++ b/docs/guides/configure_rules_via_sidecar.markdown
@@ -10,10 +10,11 @@ Cortex can be configured to use a sidecar container in the Ruler and AlertManage
 Put ConfigMaps into the specified namespace, and they are automatically detected and added as files to the Ruler and/or AlertManager containers, both of which are polling for changes on the filesystem and will make the new configurations go live dynamically.
 This feature is disabled by default. Here is a simple example:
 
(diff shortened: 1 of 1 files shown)


docs/guides/configure_rules_via_sidecar.markdown
```diff
@@ -10,10 +10,13 @@ Cortex can be configured to use a sidecar container in the Ruler and AlertManage
 Put ConfigMaps into the specified namespace, and they are automatically detected and added as files to the Ruler and/or AlertManager containers, both of which are polling for changes on the filesystem and will make the new configurations go live dynamically.
 This feature is disabled by default. Here is a simple example:
 
-*Please not that this is only supported with the local or filesystem backend. Otherwise cortex will overwrite what the operator puts in the folder with what is in s3/gcs/azure/swift. Cortex does not 2-way sync the files*
-
+Please not that this is only supported with the **local** backend. Otherwise cortex will overwrite what the operator puts in the folder with what is in s3/gcs/azure/swift. Cortex does not 2-way sync the files. Also please be aware of the following:
```
kd7lxl, line 13:
```suggestion
Please note that this is only supported with the **local** backend. Otherwise cortex will overwrite what the operator puts in the folder with what is in s3/gcs/azure/swift. Cortex does not 2-way sync the files. Also please be aware of the following:
```

docs/guides/configure_rules_via_sidecar.markdown
```diff
@@ -10,10 +10,13 @@ Cortex can be configured to use a sidecar container in the Ruler and AlertManage
...
+- local alertmanager config storage is read-only (cortex API is severely limited to read-only requests)
+- local alertmanager storage does not support state persistency (however your configurations are saved in etcd via kubernetes)
+- local ruler storage does not support RuleGroups
+- local ruler storage does not support namespaces
 ```yaml
-backend: "filesystem"
+backend: "local"
```
kd7lxl, line 19: To make it easier not to misconfigure, the chart could enforce this. Something like this:
https://github.com/cortexproject/cortex-helm-chart/blob/54ee3d99c1dff767315d63852ad9d5321eadaa88/templates/alertmanager/alertmanager-dep.yaml#L50
```yaml
        {{- if .Values.alertmanager.sidecar.enabled }}
        {{- if ne .Values.config.alertmanager_storage "local" }}
        {{ fail "sidecar only works with local storage" }}
        {{- end }}
        - name: {{ template "cortex.name" . }}-sc-alertmanager
```

nschad: ah cool. the more you know, didn't know that fail was a thing

kd7lxl: If you don't want to change any logic in this PR, I'm happy to approve once the typo above is taken care of. The logic can wait for another PR.

nschad: > If you don't want to change any logic in this PR, I'm happy to approve once the typo above is taken care of. The logic can wait for another PR.

I don't know about putting that in the already converluded yaml file. Maybe in _helpers.tpl? Or just as a warning? I rather not error-out the chart because maybe behaviour will change.

kd7lxl (approved):

PR #311, opened 2022-02-01
update maintainer info

nschad:

**What this PR does**:

Fixes #<issue number>

**Checklist**

### DIFF
--- a/docs/index.yaml
+++ b/docs/index.yaml
@@ -3,7 +3,7 @@ entries:
   cortex:
   - apiVersion: v2
     appVersion: v1.11.0
-    created: "2021-12-29T12:32:31.503943634+01:00"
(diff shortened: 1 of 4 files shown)


docs/index.yaml
```diff
@@ -308,7 +308,7 @@ entries:
     version: 0.7.0
   - apiVersion: v2
     appVersion: v1.9.0
-    created: "2021-12-29T12:32:31.471840052+01:00"
+    created: "2022-02-01T12:14:55.490871027+01:00"
```
kd7lxl, line 311: Seems like old versions should not be modified 🤔

nschad: this is the default behaviour. the date is always updated 🤷

kd7lxl (approved):

PR #314, opened 2022-02-02
don't run ct test when nothing has changed

nschad:

**What this PR does**:
Prevents the pipeline from running chart testing test for non chart related changes

Fixes #<issue number>

**Checklist**

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -29,6 +29,14 @@ jobs:
         with:
           version: v3.4.0
 
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(ct list-changed)
+          if [[ -n "$changed" ]]; then
+            echo "::set-output name=changed::true"
+          fi
+
       - name: Run chart-testing (lint)
         run: ct lint --config ct.yaml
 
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-test.yml
```diff
@@ -39,8 +47,11 @@ jobs:
 
       - name: Create kind cluster
         uses: helm/kind-action@v1.2.0
+        if: steps.list-changed.outputs.changed == 'true'
 
       - name: Run chart-testing (install)
         run: |
           kubectl create namespace cortex
           ct install --config ct.yaml --namespace cortex
+        if: steps.list-changed.outputs.changed == 'true'
```
kd7lxl, line 56: The example https://github.com/helm/chart-testing-action#example-workflow does not have this conditional. I guess if `ct install` doesn't detect a diff it is expected to exit successfully early here?

nschad: :shrug: I don't see how. It's the same command as before why would it suddenly check for diffs?

kd7lxl: If `ct list-changed` can identify what changed, that logic can definitely be called by `ct install` as well. I haven't checked ct code to verify.

I see loads of examples both ways in https://cs.github.com/?scopeName=All+repos&scope=&q=%22if%3A+steps.list-changed.outputs.changed%22+path%3A.github%2Fworkflows so it must not matter too much.

kd7lxl (approved):

PR #315, opened 2022-02-02
fix potential issue with unix permissions when using the sidecar

nschad:

**What this PR does**:

**EXPERIMENTAL**

Had some unix Issues when testing with the sidecar due to the image using 65534:65534 and cortex using root:root per default. Changes theoretically should be fine but I haven't verified it yet

Fixes #<issue number>

**Checklist**


nschad:
with the default values of cortex / sidecar this is not needed as cortex is running as root and can always read from sidecar's directory. However if you were to run cortex as USER XYZ then you would run into problems. But that's out of scope for now

PR #319, opened 2022-02-08
Enabled flags

Ziul:
**What this PR does**: Includes `enabled` flags for distributor, ingester, querier, query-frontend and store-gateway.

This PR enables the user to choose each part of the cortex gonna be released. Useful for situations where a namespace quota would request the cortex to be split into many namespaces.

**Checklist**

### DIFF
--- a/templates/ingester/ingester-hpa.yaml
+++ b/templates/ingester/ingester-hpa.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.ingester.enabled -}}
 {{- with .Values.ingester.autoscaling -}}
 {{- if .enabled }}
 apiVersion: autoscaling/v2beta2
@@ -25,5 +26,6 @@ spec:
   behavior:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+{{- end }}
 {{- end }}
 {{- end }}
--- a/templates/querier/querier-hpa.yaml
+++ b/templates/querier/querier-hpa.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.querier.enabled -}}
 {{- with .Values.querier.autoscaling -}}
 {{- if .enabled }}
 apiVersion: autoscaling/v2beta2
@@ -35,5 +36,6 @@ spec:
   behavior:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+{{- end }}
 {{- end }}
 {{- end }}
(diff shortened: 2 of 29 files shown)


templates/ingester/ingester-hpa.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.ingester.enabled -}}
```
nschad, line 1: Duplicate. Line 2 already does that

templates/querier/querier-hpa.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.querier.enabled -}}
```
nschad, line 1: duplicate `with` is sufficient

nschad (changes requested):
Thanks! Please add a CHANGELOG entry

templates/ingester/ingester-hpa.yaml
Ziul: Line 2 enables the `.Values.ingester.autoscaling`. Without the `.Values.ingester.enabled` check, would be possible to create a hpa without the ingester deployment.
Replace this `if` with `if and  .Values.ingester.enabled .Values.ingester.autoscaling.enabled` and removing the second line would be a better choice, don’t you think?

templates/querier/querier-hpa.yaml
Ziul: I believe the same solution from ingester-hpa would work here.

templates/ingester/ingester-hpa.yaml
nschad: Yeah okay if you want to be explicit about it than we can do your suggestion

templates/querier/querier-hpa.yaml
nschad: yup same as above :+1:

templates/ingester/ingester-hpa.yaml
Ziul: Commit done!

nschad (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * [ENHANCEMENT] Fix the indentation of memcached guide #309
 * [ENHANCEMENT] Added api endpoints for Grafana 8 unified alerting #291
 * [ENHANCEMENT] Add guide on how to configure alertmanager/ruler with configmap #313
+* [ENHANCEMENT] Includes enable flags for each component
```
nschad, line 16:
```suggestion
* [ENHANCEMENT] Includes enable flags for each component #319
```

nschad (changes requested):

nschad (approved):

Ziul:
I amended the last commit to include my signature, so the DCO bot stop failing.

kd7lxl (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Includes enable flags for each component #319
```
nschad, line 5: Ooops missed that one :) but now it's correct.

PR #227, opened 2021-09-22
Release 1.0.0

ShuzZzle:
### **Release 1.0.0**

- Allow overriding cortex Version
- Update Policy API Version
- Deprecate *.persistence
- Remove clusterPort for alertmanager
- Remove legacy alertmanager clustering
- Removed unused `nginx.serviceMonitor` section
- Added IngressClass to ingressResource
- Update default values for bitnami memcached chart
- Drop default chunk storage **config** (the chart still fully supports chunk storage if you so choose too)
- server.grpc_server_max_recv_msg_size: 104857600 -> **10485760**
- server.grpc_server_max_send_msg_size: 104857600 -> **10485760**
- server.grpc_server_max_concurrent_streams: 1000 -> **10000**
- server.ingester_client.grpc_client_config.max_recv_msg_size: 104857600 -> **10485760**
- server.ingester_client.grpc_client_config.max_send_msg_size: 104857600 -> **10485760**
- config.limits.enforce_metric_name: false -> **true**
- config.ruler.enable_api: false -> **true**
- configure memberlist automatically
- config.ingester.lifecycler.final_sleep: 0s -> **30s**

**Which issue(s) this PR fixes**:
Fixes #202, #252

**Checklist**
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #321, opened 2022-02-16
Fix permissions on AlertManager sidecar

gburton1:
**What this PR does**:
This restores the user that the alertmanager sidecar runs as to 0. It had been changed during this refactor in https://github.com/cortexproject/cortex-helm-chart/pull/227/files for the 1.0.0 release:
```
-    securityContext:
-      runAsUser: 0
+    containerSecurityContext:
+      enabled: true
+      readOnlyRootFilesystem: true
```

Fixes https://cloud-native.slack.com/archives/CCYDASBLP/p1642492446058600

**Checklist**
- [x ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## master / unreleased
+* [BUGFIX] AlertManager sidecar was changed to not run as user 0 which broke the sidecar
 
 ## 1.3.0 / 2022-02-10
 
--- a/values.yaml
+++ b/values.yaml
@@ -323,6 +323,7 @@ alertmanager:
     resource: "both"
     containerSecurityContext:
       enabled: true
+      runAsUser: 0
       readOnlyRootFilesystem: true
 
 distributor:


nschad (commented):
It's not necessary if cortex is running as root. I wanted to do the same thing but it works like this out of the box See: #315

nschad:
In my deployment we have the following:

* sidecar runs as configured nobody:nobody
* cortex runs as root:root
```
~ $ ls -alh /data/
total 24K    
drwxrwxrwx    3 root     root        4.0K Feb 16 14:50 .
drwxr-xr-x    1 root     root        4.0K Feb 16 14:50 ..
drwxr-xr-x    2 nobody   nobody      4.0K Feb 16 21:20 fake
~ $ ls -alh /data/fake
total 36K    
drwxr-xr-x    2 nobody   nobody      4.0K Feb 16 21:20 .
drwxrwxrwx    3 root     root        4.0K Feb 16 14:50 ..
-rw-r--r--    1 nobody   nobody      1.4K Feb 16 14:50 fake.yaml
-rw-r--r--    1 root     root         589 Feb 16 21:20 notifications
-rw-r--r--    1 root     root           0 Feb 16 21:20 silences
```
so cortex actually has more permission thus no error. I do understand if it were the other way around tho. If cortex were to create the `fake` folder with root:root then yeah it would not work

gburton1:
Hmm. This means that the change was backwards incompatible to past live deployments of this chart, unless action is taken to blow away existing files in all environments, since the existing files were created under a past version of Cortex chart when the sidecar was user 0, and therefore maintain that older ownership.

Although this doesn't quite explain the experience of Charlie on the thread, since he was creating a new deployment from scratch. Unless maybe he had started on an earlier version too.

nschad:
> Hmm. This means that the change was backwards incompatible to past live deployments of this chart, unless action is taken to blow away existing files in all environments, since the existing files were created under a past version of Cortex chart when the sidecar was user 0, and therefore maintain that older ownership.

Yeah okay that makes sense. Looks like we missed that one :(. To be fair it was a major release, haha :D

gburton1:
Not sure why the linter check is unhappy

nschad:
> Not sure why the linter check is unhappy

Run `helm-docs` in the repo to update the README.md https://github.com/norwoodj/helm-docs. But I don't if we will acccept this PR. Preferably we don't want anything to run as root.

disambiguationuk:
Interesting, the only way we can get this to work is to use this `runAsUser: 0` override too.

PR #330, opened 2022-03-07
unable nginx keepalive to cortex components

jmcarp:
Signed-off-by: Josh Carp <jm.carp@gmail.com>

**What this PR does**:

Add a small nginx optimization: enable keepalive to cortex components. Note that this requires configuring upstreams with explicit `upstream` blocks. I noticed this in a post from nginx at https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#no-keepalives; I tested it out and got a modest drop in nginx cpu use.

**Checklist**

### DIFF
--- a/templates/nginx/nginx-config.yaml
+++ b/templates/nginx/nginx-config.yaml
@@ -40,35 +40,35 @@ data:
       {{- if .Values.alertmanager.enabled }}
       upstream alertmanager {
         server {{ template "cortex.fullname" . }}-alertmanager.{{ $rootDomain }};
-        keepalive {{ .Values.nginx.config.upstream_keepalive }};
+        keepalive_timeout 0;
       }
       {{- end }}
 
       {{- if .Values.configs.enabled }}
       upstream configs {
         server {{ template "cortex.fullname" . }}-configs.{{ $rootDomain }};
-        keepalive {{ .Values.nginx.config.upstream_keepalive }};
+        keepalive_timeout 0;
       }
       {{- end }}
 
       {{- if .Values.distributor.enabled }}
       upstream distributor {
         server {{ template "cortex.fullname" . }}-distributor.{{ $rootDomain }};
-        keepalive {{ .Values.nginx.config.upstream_keepalive }};
+        keepalive_timeout 0;
       }
       {{- end }}
 
       {{- if .Values.query_frontend.enabled }}
       upstream query-frontend {
         server {{ template "cortex.fullname" . }}-query-frontend.{{ $rootDomain }};
-        keepalive {{ .Values.nginx.config.upstream_keepalive }};
+        keepalive_timeout 0;
       }
       {{- end }}
 
       {{- if .Values.ruler.enabled }}
       upstream ruler {
         server {{ template "cortex.fullname" . }}-ruler.{{ $rootDomain }};
-        keepalive {{ .Values.nginx.config.upstream_keepalive }};
+        keepalive_timeout 0;
       }
       {{- end }}
 
@@ -78,7 +78,6 @@ data:
         proxy_send_timeout 300s;
         proxy_read_timeout 300s;
         proxy_http_version 1.1;
-        proxy_set_header Connection "";  ## Enable keepalive to upstreams
 
         {{- range $key, $value := .Values.nginx.config.setHeaders }}
         proxy_set_header {{ $key }} {{ $value }};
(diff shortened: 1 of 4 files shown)


templates/nginx/nginx-config.yaml
```diff
@@ -37,12 +37,48 @@ data:
       {{ tpl . $ | nindent 6 }}
       {{- end }}
 
+      {{- if .Values.alertmanager.enabled }}
+      upstream alertmanager {
+        server {{ template "cortex.fullname" . }}-alertmanager.{{ $rootDomain }};
+        keepalive 32;
```
nschad, line 43: hmm do you see a reason of having this configurable?

jmcarp: I haven't tried to tune keepalives extensively, but maybe it would be useful to make configurable. Updated the patch to do this.

nschad (approved):

kd7lxl (approved):

nschad:
What about load-balancing @jmcarp? Wouldn't the keep-alive blocks cause problems and lead to further uneven load distribution on specifically the distributors?

I can already see that happening on my distributors using the default helm-chart "flow" (Internet -> Ingress -> nginx -> distributor)

image

templates/nginx/nginx-config.yaml
```diff
@@ -37,12 +37,48 @@ data:
...
+      {{- if .Values.configs.enabled }}
+      upstream configs {
+        server {{ template "cortex.fullname" . }}-configs.{{ $rootDomain }};
+        keepalive {{ .Values.nginx.config.upstream_keepalive }};
+      }
+      {{- end }}
+
+      {{- if .Values.distributor.enabled }}
+      upstream distributor {
+        server {{ template "cortex.fullname" . }}-distributor.{{ $rootDomain }};
```
kd7lxl, line 56: Based on the definition here: http://nginx.org/en/docs/http/ngx_http_upstream_module.html#server
> A domain name that resolves to several IP addresses defines multiple servers at once.

I think it would be better to use the `distributor-headless` service. We should also use `resolve` to monitor for changes in the set of pods.
> `resolve` monitors changes of the IP addresses that correspond to a domain name of the server, and automatically modifies the upstream configuration without the need of restarting nginx
```suggestion
        server {{ template "cortex.fullname" . }}-distributor-headless.{{ $rootDomain }} resolve;
```

nschad: I guess we should do that for every service then that has a headless service.

jmcarp: The `resolve` option is only supported for nginx commercial, so not sure we can do this.

nschad: > > The `resolve` option is only supported for nginx commercial, so not sure we can do this.
> 
> Oh, maybe that's why ingress-nginx uses lua to resolve pod addresses. This is convincing me that the nginx gateway in this chart isn't very useful and most use cases might be better served by Ingress.

Agree. However we would have to abandon some features

kd7lxl: Okay, so without `resolve`, how can we distribute requests... Is the design assumption that the `keepalive 16` will result in open connections roughly balanced to each pod? And nginx will round-robin requests to those connections?

BTW, here I use linkerd to balance load to the distributors. It works quite well for that, but I wouldn't expect everyone to deploy linkerd.

nschad: > Okay, so without `resolve`, how can we distribute requests... Is the design assumption that the `keepalive 16` will result in open connections roughly balanced to each pod? And nginx will round-robin requests to those connections?

I would assume that we actually have to do the opposite and ensure that we do not use any kind of keepalive connection to achieve fair distribution.

https://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_timeout setting this to `0` should disable that completely

> The `resolve` option is only supported for nginx commercial, so not sure we can do this.

Where does it say that?

nschad:
@jmcarp Hey Josh,

I did some experiencing and setting keepalive for cortex is actually in my test very problematic as connections are kept and therefore load is distributed unevenly. **But** the idea with the individual upstreams had/has a very positive effect in my tests as it seems that nginx is now forced to always resolve said upstream server

image

jmcarp:
Thanks for fixing up the PR @nschad! Just curious, what are you plotting in that graph? I'm actually a little surprised that moving configurations to upstream blocks without keepalives would help, but great if it does.

nschad:
Closed in favor of #366 . Thank you so much for contribution!

PR #339, opened 2022-03-26
feat(nginx): ensure dns resolver respects a ttl

cassandracomar:
when services behind the nginx proxy roll or scale up, nginx does not pick
up the new ips as it never rechecks DNS until it is restarted. this can
be mitigated by setting a ttl with the valid parameter on the DNS
resolver AND by ensuring the domain lookups are made through a variable.

see: https://serverfault.com/questions/240476/how-to-force-nginx-to-resolve-dns-of-a-dynamic-hostname-everytime-when-doing-p/593003#593003

**What this PR does**:

ensure the resolver has a suitable validity period and move domains
behind an nginx variable, forcing nginx to recheck DNS when proxying
requests.

Fixes #<issue number>

**Checklist**

### DIFF
--- a/templates/nginx/nginx-config.yaml
+++ b/templates/nginx/nginx-config.yaml
@@ -31,7 +31,7 @@ data:
       access_log   /dev/stderr  main;
       sendfile     on;
       tcp_nopush   on;
-      resolver {{ default (printf "kube-dns.kube-system.svc.%s" .Values.clusterDomain ) .Values.nginx.config.dnsResolver }};
+      resolver {{ default (printf "kube-dns.kube-system.svc.%s" .Values.clusterDomain ) .Values.nginx.config.dnsResolver }} valid={{ default "10s" .Values.nginx.config.dnsTtl }};
 
       {{- with .Values.nginx.config.httpSnippet }}
       {{ tpl . $ | nindent 6 }}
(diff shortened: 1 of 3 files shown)


templates/nginx/nginx-config.yaml
```diff
@@ -68,98 +68,103 @@ data:
         }
 
         {{- if .Values.distributor.enabled }}
+        set $distributor_backend "http://{{ template "cortex.fullname" . }}-distributor.{{ $rootDomain }}";
```
nschad, line 71: ah I see this is much like #330. I think I would prefer the upstream blocks instead of just a variable tho. Maybe we could allow the user to configure said upstream block as well

```yaml
      {{- if .Values.alertmanager.enabled }}
      upstream alertmanager {
        server {{ template "cortex.fullname" . }}-alertmanager.{{ $rootDomain }};
        {{- with .Values.nginx.config.alertmanager }}
        {{ tpl . $ | nindent 4 }}
        {{- end }}
      }
      {{- end }}
      ```
      
      Or maybe this is a bit overkill. Thoughts @kd7lxl ?

Comment (cassandracomar) [reply templates/nginx/nginx-config.yaml:71]: the upstream block prevents setting the TTL on the dns cache, unfortunately.

Comment (nschad) [reply templates/nginx/nginx-config.yaml:71]: Do you have thoughts on this? @kd7lxl

File: templates/nginx/nginx-config.yaml [hunk 2]
```diff
@@ -31,7 +31,7 @@ data:
       access_log   /dev/stderr  main;
       sendfile     on;
       tcp_nopush   on;
-      resolver {{ default (printf "kube-dns.kube-system.svc.%s" .Values.clusterDomain ) .Values.nginx.config.dnsResolver }};
+      resolver {{ default (printf "kube-dns.kube-system.svc.%s" .Values.clusterDomain ) .Values.nginx.config.dnsResolver }} valid={{ default "10s" .Values.nginx.config.dnsTtl }};
```
kd7lxl, line 34: The solution seems sound to me. I have not exercised this feature of nginx but I trust your testing.

I don't think we need double-defined defaults here, since you already define a default in the values.yaml.
```suggestion
      resolver {{ default (printf "kube-dns.kube-system.svc.%s" .Values.clusterDomain ) .Values.nginx.config.dnsResolver }}{{ with .Values.nginx.config.dnsTTL }} valid={{ . }}{{ end }};
```

values.yaml
```diff
@@ -1151,6 +1151,7 @@ nginx:
   http_listen_port: 80
   config:
     dnsResolver: kube-dns.kube-system.svc.cluster.local
+    dnsTtl: 10s
```
kd7lxl, line 1154: This is a nitpick since I acknowledge we already have [multiple](https://github.com/cortexproject/cortex-helm-chart/blob/c7a7f4e198c19a9679504ea062c61795fba765ef/values.yaml#L382) [styles](https://github.com/cortexproject/cortex-helm-chart/blob/c7a7f4e198c19a9679504ea062c61795fba765ef/values.yaml#L298) in this project, but I prefer the style of capitalizing acronyms.
```suggestion
    dnsTTL: 10s
```

kd7lxl (commented):
Thanks for your contribution and sorry for the indecisiveness. The solution in this PR is sound, but I've already approved #366 so it might not be worth your effort to continue with this PR.

cassandracomar:
@kd7lxl issue is, if the upstream block is merged, we have no way to force nginx to re-resolve DNS. as outlined in the stack overflow thread I linked (https://serverfault.com/questions/240476/how-to-force-nginx-to-resolve-dns-of-a-dynamic-hostname-everytime-when-doing-p/593003#593003), nginx will not respect the resolver ttl. the only way to make nginx not attempt to send traffic to pods that don't exist or to send traffic to new pods it's not aware of is to restart/reload it. this makes autoscaling cortex distributors impossible without an additional controller to acknowledge the change to the running distributors and restart the nginx pods.

for a time, we also tried bypassing nginx entirely and pointing prometheus directly at the distributors, but this fails because each prometheus process establishes a connection to a single distributor and forwards all samples to that single distributor.

I'm not sure how anyone is able to make this work at scale without either patching this chart or simply disabling autoscaling on cortex distributors.

nschad:
Closed in favor of #366 . Thank you so much for contribution!

PR #343, opened 2022-04-18
Add topology spread constrant option

locmai:
**What this PR does**:

Provide the ability to configure Pod Topology Spread Constraints.

**Checklist**

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -197,6 +197,7 @@ alertmanager:
     prometheus.io/port: '8080'
 
   nodeSelector: {}
+  topologySpreadConstraints: []
   affinity: {}
   annotations: {}
 
@@ -359,6 +360,7 @@ distributor:
     prometheus.io/port: '8080'
 
   nodeSelector: {}
+  topologySpreadConstraints: []
   affinity:
     podAntiAffinity:
       preferredDuringSchedulingIgnoredDuringExecution:
@@ -468,6 +470,7 @@ ingester:
     prometheus.io/port: '8080'
 
   nodeSelector: {}
+  topologySpreadConstraints: []
   affinity:
     podAntiAffinity:
       preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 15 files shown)


values.yaml
```diff
@@ -468,6 +470,7 @@ ingester:
     prometheus.io/port: '8080'
 
   nodeSelector: {}
+  topologySpreadConstraints: []
```
gburton1, line 473: Question for reviewers: given that ingesters have a pod disruption budget of 1, would it make sense to revise the default podAntiAffinity config a few lines below this to make it a hard requirement that more than one ingester should never be scheduled to the same node? That behavior could also be implemented using topologySpreadConstraints constructs, which would eliminate the need for the podAntiAffinity config.

This is desirable because presently it is possible to get two pods on one node, and then when node patching happens, the process gets stuck because it can't evict two ingester pods in order to patch the node.

nschad (commented):
argh nice but you have to use helm-docs@1.8.1. apparently they changed some things

https://github.com/norwoodj/helm-docs/releases/tag/v1.8.1

locmai:
@nschad Thanks for taking a look, I just pushed a new one to solve that.

nschad (approved):

values.yaml
nschad: would you be open to create an seperate PR after this is merged?

gburton1: Yes!

kd7lxl (approved):
Sorry, didn't realize you were waiting on me. Looks good.

values.yaml
kd7lxl: > would it make sense to revise the default podAntiAffinity config a few lines below this to make it a hard requirement that more than one ingester should never be scheduled to the same node?

We do this for our production deployments with an override. We should probably pick a clear strategy for this chart on whether the chart defaults represent good production values or whether they are more tuned for testing. It is helpful to be able to deploy multiple ingester pods to a single node for testing.

PR #366, opened 2022-06-08
feat: add validDns option and headless distributor to Nginx

locmai:
Signed-off-by: Loc Mai <lmai@axon.com>

**What this PR does**:

Address the NGINX issue on Kubernetes when rolling restart the distributor pods would cause an amount of ingestion failures on the write path.

This is more of a Kubernetes issue but we could have this as a work around. See https://github.com/kubernetes/kubernetes/issues/24092

No issue created.

## Template results

```yaml
# values.yaml
nginx:
  config:
    dnsValid: 15s
    distributorHeadless: true
```

nginx-config - configmap:

```
...
        resolver kube-dns.kube-system.svc.cluster.local 15s;

        # Distributor Config
        location = /ring {
          proxy_pass      http://RELEASE-NAME-cortex-distributor-headless.default.svc.cluster.local:8080$request_uri;
        }
```

**Checklist**

cc @juan-ramirez-sp @gburton1


nschad (commented):
Ah this is very similiar to #339 and probably also to #330. Quiet frankly I not sure which road we should go. However I do see the necessity

locmai:
It is indeed. I think @juan-ramirez-sp tried the #330 but didn't work, see this Slack discussion https://cloud-native.slack.com/archives/CCYDASBLP/p1654268570724589 . 

We tested this change on our deployment and it works nicely. Me and @gburton1 could take a look at the #339 though, I think it's pretty similar.

nschad:
> It is indeed. I think @juan-ramirez-sp tried the #330 but didn't work, see this Slack discussion https://cloud-native.slack.com/archives/CCYDASBLP/p1654268570724589 .
> 
> We tested this change on our deployment and it works nicely. Me and @gburton1 could take a look at the #339 though, I think it's pretty similar.

Ah okay upon further investigation: https://serverfault.com/questions/240476/how-to-force-nginx-to-resolve-dns-of-a-dynamic-hostname-everytime-when-doing-p this seems to be a ongoing issue for a lot of people. 

The best solution might be this or #339 which is effectively the same

values.yaml
```diff
@@ -1276,7 +1276,10 @@ nginx:
     #   setHeaders:
     #     X-Scope-OrgID: $remote_user
     basicAuthSecretName: ""
-
+    # -- (optional) Including the valid parameter to the `resolver` directive to re-resolve names every `dnsValid` seconds/minutes
+    dnsValid: ""
```
kd7lxl, line 1280: Since this is broken now, I would be happy to see a default that behaves nicely.
```suggestion
    dnsValid: "15s"
```

kd7lxl (approved):

templates/nginx/nginx-config.yaml
```diff
@@ -71,7 +71,7 @@ data:
 
         # Distributor Config
         location = /ring {
-          proxy_pass      http://{{ template "cortex.fullname" . }}-distributor.{{ $rootDomain }}$request_uri;
+          proxy_pass      http://{{ template "cortex.fullname" . }}-distributor{{- if .Values.nginx.config.distributorHeadless }}-headless{{- end }}.{{ $rootDomain }}$request_uri;
```
gburton1, line 74: I think `headless` should be applied to all distributor endpoints below also.

values.yaml
nschad: good idea

values.yaml
```diff
@@ -1276,7 +1276,10 @@ nginx:
...
+    # -- (optional) If true Nginx will use the distributor headless service for `/ring` path. See https://github.com/kubernetes/kubernetes/issues/24092
+    distributorHeadless: false
```
nschad, line 1282: Like @gburton1 suggested if we replace everything with the headless service we don't need that config option here

values.yaml
locmai: Yeah great idea. Anyone wanna change the option name? I see in the other PR they have dnsTTL. 

By the way, after I re-tested the whole case here, it looks like the `valid=15s` is all we need to solve the problem (no headless service change required).

templates/nginx/nginx-config.yaml
nschad: have you tried this? with headless you get multiple dns results with the pod individual ip address. Does "load-balancing" still work?

values.yaml
nschad: yeah I like dnsTTL more

nschad:
but @locmai I don't quite understand how `ClusterIP` is a problem. The virtual IP of the distributor service doesn't change when pods are scaling up/down. kube-proxy then takes over routing that traffic to any distributor pod (fake round-robin or actual round-robin with ipvs backend)

locmai:
> but @locmai I don't quite understand how `ClusterIP` is a problem. The virtual IP of the distributor service doesn't change when pods are scaling up/down. kube-proxy then takes over routing that traffic to any distributor pod (fake round-robin or actual round-robin with ipvs backend)

Yeah I think it's not. As I mentioned in the comment above, I tested again and the `valid=15s` is all we need. My guess is that after nginx resolve the DNS into the service, then retrieve the list of endpoints (Pods' IPs) then it would cache it right there, so by having the small TTL, it would just invalidate the old record and do the resolving process all again.

This may put some more load on the nginx though.

nschad:
> > but @locmai I don't quite understand how `ClusterIP` is a problem. The virtual IP of the distributor service doesn't change when pods are scaling up/down. kube-proxy then takes over routing that traffic to any distributor pod (fake round-robin or actual round-robin with ipvs backend)
> 
> Yeah I think it's not. As I mentioned in the comment above, I tested again and the `valid=15s` is all we need. My guess is that after nginx resolve the DNS into the service, then retrieve the list of endpoints (Pods' IPs) then it would cache it right there, so by having the small TTL, it would just invalidate the old record and do the resolving process all again.
> 
> This may put some more load on the nginx though.

Do you just wanna go ahead with only the `dnsTTL` change and keep service as is?

locmai:
Yeah I just updated the PR @nschad @kd7lxl

nschad (approved):

locmai:
> Error: timed out waiting for the condition
> [41](https://github.com/cortexproject/cortex-helm-chart/runs/6827568562?check_suite_focus=true#step:9:42)

Sounds like a hiccup/flaky test @@

nschad:
> > Error: timed out waiting for the condition
> > [41](https://github.com/cortexproject/cortex-helm-chart/runs/6827568562?check_suite_focus=true#step:9:42)
> 
> Sounds like a hiccup/flaky test @@

> 2022/06/10 08:38:29 [emerg] 1#1: host not found in resolver "15s" in /etc/nginx/nginx.conf:21

locmai:
oh shoot, got it now :D

nschad:
@locmai I think the nginx config is wrong. How does that work for you o.O

Should be
> resolver 127.0.0.1 [::1]:5353 valid=30s;
http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver

README.md
```diff
@@ -489,6 +489,7 @@ Kubernetes: `^1.19.0-0`
...
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;client_max_body_size | string | `"1M"` | ref: http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size |
 | nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;dnsResolver | string | `"kube-dns.kube-system.svc.cluster.local"` |  |
+| nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;dnsTTL | string | `"15s"` | (optional) Including the valid parameter to the `resolver` directive to re-resolve names every `dnsValid` seconds/minutes |
```
locmai, line 492:
```suggestion
| nginx.&ZeroWidthSpace;config.&ZeroWidthSpace;dnsTTL | string | `"15s"` | (optional) Including the valid parameter to the `resolver` directive to re-resolve names every `dnsTTL` seconds/minutes |
```

locmai: have to fix this also

locmai:
@nschad sorry my bad. Templated out here:

```
resolver kube-dns.kube-system.svc.cluster.local valid=15s;
```
missed the `valid=`. I tested my local with simple editing the configmap.

Issue #378, opened 2022-07-11
Add option for running in "monolithic" (all) mode


TaylorMutch:
I would love to be able to run Cortex in smaller environments in a "monolithic" mode, so I can stand up integration tests without having to specify an entire distributed stack.

We could add the option to specify a "mode", and keep the current distributed mode as the default.

TaylorMutch:
I'm still interested in contributing this

PR #379, opened 2022-07-11
Add monolithic mode options

TaylorMutch:
**What this PR does**:

Adds the ability to deploy the Cortex helm chart  in a "monolithic" mode

Fixes https://github.com/cortexproject/cortex-helm-chart/issues/378

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [BUGFIX] Fix nginx extraContainers to indent correctly #373
 * [DEPENDENCY] Update dependency nginx to v1.23 #375
 * [DEPENDENCY] Update dependency kiwigrid/k8s-sidecar to v1.19.2 #370
+* [FEATURE] Add monolithic mode options #379
 
 ## 1.6.0 / 2022-06-10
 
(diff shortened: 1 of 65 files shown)


kd7lxl (commented):
What you have prepared looks great and should work fine.

I am conflicted on whether this should just be its own chart. It disables almost the whole chart, so maybe it would be simpler to just have two variations of the chart, one for distributed and one for monolith.

kd7lxl (changes requested):
I've thought about it a bit more and with the common configmap, I do see the value in including monolithic mode in the same chart. Some things I would like to see in this PR:
1. Consolidate the pod template to be common in the `Deployment` and `Statefulset` (or help me understand the value in maintaining them separately -- I'm trying to keep it DRY)
2. Add a section to the main part of the readme explaining monolithic mode and when you might use it, so it's more obvious to users that it's available.
3. Add test values for it in `ci/`

TaylorMutch:
> 1. Consolidate the pod template to be common in the `Deployment` and `Statefulset` (or help me understand the value in maintaining them separately -- I'm trying to keep it DRY)

I think a pod template that is shared makes sense; I've seen that implemented elsewhere as well. I followed the pattern for the `ingester` templates which has them separated. I can implement a shared pod template for the monolith, and if we like that we could do that in the rest of the chart in a separate PR. What do you think?

kd7lxl:
> if we like that we could do that in the rest of the chart in a separate PR. What do you think?

That would be nice, but it will depend on whether it can be done in a backwards compatible way. I haven't looked closely at the difference between the ingester deployment and statefulset to know yet if they could be updated to the same template without breaking anything. If something like the selector labels don't match it could present a problem.

PR #412, opened 2022-11-08
fix: nginx fails to forward requests to new pods

TuTuRyYE:
Signed-off-by: ajaumet <ajaumet@scaleway.com>

**What this PR does**:
This PR fixes NGINX not being able to reach newly created pods after deployment if NGINX is not reloaded in the process.

Fixes #<issue number>

**Checklist**
- [ x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`


nschad:
Doesn't nginx re-resolve dns names, now? I believe we added this in #366. What you are doing is an "hack" so that nginx always re-resolves the dns name since the information is stored in variable, right?

TuTuRyYE:
After some checks on our side, this seems to be resolved. I'm closing this PR.

PR #414, opened 2022-12-05
Start using autoscaling/v2 for HorizontalPodAutoscaler

GuyKomari:
What this PR does / why we need it:

Problem: chart-testing is failing for:
"autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2"

https://github.com/cortexproject/cortex-helm-chart/actions/runs/3606103570/jobs/6077123598

Solution: update charts to start using autoscaler/v2

Signed-off-by: guykomari <guykomari@gmail.com>

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -153,3 +153,14 @@ Get volume of config secret of configMap
     secretName: {{ template "cortex.fullname" . }}
   {{- end }}
 {{- end -}}
+
+{{/*
+Get cortex hpa version by k8s version
+*/}}
+{{- define "cortex.hpaVersion" -}}
+{{- if or (.Capabilities.APIVersions.Has "autoscaling/v2/HorizontalPodAutoscaler") (semverCompare ">=1.23" .Capabilities.KubeVersion.Version) -}}
+autoscaling/v2
+{{- else -}}
+autoscaling/v2beta2
+{{- end -}}
+{{- end -}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Start using autoscaling/v2 for HorizontalPodAutoscaler in v1.23+ #414
 * [FEATURE] add purger components to cortex #407
 * [ENHANCEMENT] Add verboseLogging option to nginx config #402
 * [DEPENDENCY] Update quay.io/cortexproject/cortex Docker tag to v1.13.1 #401
--- a/templates/distributor/distributor-hpa.yaml
+++ b/templates/distributor/distributor-hpa.yaml
@@ -1,6 +1,6 @@
 {{- if and .Values.distributor.enabled .Values.distributor.autoscaling.enabled }}
 {{- with .Values.distributor.autoscaling -}}
-apiVersion: autoscaling/v2beta2
+apiVersion: {{ include "cortex.hpaVersion" $ }}
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ include "cortex.distributorFullname" $ }}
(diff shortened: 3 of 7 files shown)


nschad (commented):
We should do something like this until 1.26+. To keep functionality with older clusters.

```yaml
{/*
Determine the policy api version
*/}}
{{- define "cortex.pdbVersion" -}}
{{- if or (.Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget") (semverCompare ">=1.21" .Capabilities.KubeVersion.Version) -}}
policy/v1
{{- else -}}
policy/v1beta1
{{- end -}}
{{- end -}}
```

nschad (changes requested):
Thank you, however please put the code in `templates/_helpers.tpl` global helpers file to avoid unnecessary code duplication just like with the pdbversion.

Then you can refer to it like this
```yaml
apiVersion: {{ include "cortex.hpaVersion" . }}
```

templates/_helpers.tpl
```diff
@@ -153,3 +153,14 @@ Get volume of config secret of configMap
...
 {{- end -}}
+
+{{/*
+Get cortex hpa version by k8s version
+*/}}
+{{- define "cortex.hpaVersion" -}}
+{{- if or (.Capabilities.APIVersions.Has "autoscaling/v2") (semverCompare ">=1.23" .Capabilities.KubeVersion.Version) -}}
+apiVersion: autoscaling/v2
+{{- else -}}
+apiVersion: autoscaling/v2beta2
```
nschad, line 164:
```suggestion
autoscaling/v2
{{- else -}}
autoscaling/v2beta2
```

nschad (changes requested):

GuyKomari:
> Thank you, however please put the code in `templates/_helpers.tpl` global helpers file to avoid unnecessary code duplication just like with the pdbversion.
> 
> Then you can refer to it like this
> 
> ```yaml
> apiVersion: {{ include "cortex.hpaVersion" . }}
> ```
> 
> _On a side note: We should also check if something changed between v1beta1 and v1 (but I guess it shouldn't have?)_

by the docs, it seems that nothing has changed.

templates/_helpers.tpl
```diff
@@ -153,3 +153,14 @@ Get volume of config secret of configMap
     secretName: {{ template "cortex.fullname" . }}
   {{- end }}
 {{- end -}}
+
+{{/*
+Get cortex hpa version by k8s version
+*/}}
+{{- define "cortex.hpaVersion" -}}
+{{- if or (.Capabilities.APIVersions.Has "autoscaling/v2") (semverCompare ">=1.23" .Capabilities.KubeVersion.Version) -}}
```
nschad, line 161:
```suggestion
{{- if or (.Capabilities.APIVersions.Has "autoscaling/v2/HorizontalPodAutoscaler") (semverCompare ">=1.23" .Capabilities.KubeVersion.Version) -}}
```

Should be this

nschad (changes requested):

nschad:
@GuyKomari Check the Lint Test Job for error logs. There is some error evaluating the helm-chart.

Also please update the `CHANGELOG.md`

GuyKomari:
> @GuyKomari Check the Lint Test Job for error logs. There is some error evaluating the helm-chart.
> 
> Also please update the `CHANGELOG.md`
> 
> * [ ]  `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

I ran the ct lint locally and it passes now, so the github action should pass now.

GuyKomari:
@nschad once we merge this pr, when do we think we can also merge https://github.com/cortexproject/cortex-helm-chart/pull/413
and create a new release?

nschad (approved):

kd7lxl (approved):
LGTM, thank you!

PR #435, opened 2023-02-16
change alertmanager-svc-headless from http to grpc port

AlexandreRoux:
**What this PR does**:
This PR correct the alertmanager-headless service with exposing grpc port instead of http.

```
cortex-alertmanager-headless     ClusterIP   None             <none>        9095/TCP    1d
cortex-distributor-headless      ClusterIP   None             <none>        9095/TCP    1d
cortex-ingester-headless         ClusterIP   None             <none>        9095/TCP    1d
cortex-query-frontend-headless   ClusterIP   None             <none>        9095/TCP    1d
cortex-store-gateway-headless    ClusterIP   None             <none>        9095/TCP    1d
```

**Which issue(s) this PR fixes**:
This PR does not fix an issue BUT is related to conversation in #420

**Checklist**
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #460, opened 2023-05-12
WIP: configure alertmanager HA cluster mode for sts

nschad:
**What this PR does**:

Add's support for automatically adding peers for alertmanager.
I've disabled alertmanager deployment mode as I don't see how we could realistically configure the cluster mode for it.
(deployment names are not predictable/consistent). But I don't have a strong opinion about it, I just think it's unneeded.

CC: @kd7lxl @AlexandreRoux 

Note: the ci test values are increased for testing purposes. They will be reverted back if we ever merge this PR

TODO:

- Revert ci/ changes
- Remove Alertmanager Deployment Mode if reasonable (clean up values and sts condition)

Fixes #<issue number>

**Checklist**


templates/alertmanager/alertmanager-statefulset.yaml
```diff
@@ -152,6 +154,12 @@ spec:
           args:
             - "-target=alertmanager"
             - "-config.file=/etc/cortex/cortex.yaml"
+            {{- if gt (int .Values.alertmanager.replicas) 1}}
+            {{- $fullName := include "cortex.alertmanagerFullname" . }}
+            {{- range $i := until (int .Values.alertmanager.replicas) }}
+            - "-alertmanager.cluster.peers={{ $fullName }}-{{ $i }}.{{ $fullName }}-headless.{{ $.Release.Namespace }}.svc.cluster.local:{{ $svcClusterPort }}"
```
nschad, line 160: Not sure if this even technically correct. Since according to the documentation this is supposed to be comma seperated list which is different from alertmanager configuration where you have to repeat the flag for each peer.

AlexandreRoux: > Not sure if this even technically correct. Since according to the documentation this is supposed to be comma seperated list which is different from alertmanager configuration where you have to repeat the flag for each peer.

I see you are referencing https://github.com/cortexproject/cortex/blob/eaf9463df107d9b555ed2acc92df54db6472edc3/integration/configs.go#L88

AlexandreRoux: > > alertmanager.cluster.peers
> 
> I don't that why I said it should be single string and comma separated (Doc: https://cortexmetrics.io/docs/configuration/configuration-file/)

I misunderstood your message so, sorry.

nschad: Uh I don't know if I can enable this after the fact. I can close this PR and you can just edit your PR #435?

AlexandreRoux: > Uh I don't know if I can enable this after the fact. I can close this PR and you can just edit your PR #435?

sounds good !

PR #467, opened 2023-06-19
Add option for configuring HPA with custom/external metrics

locmai:
**What this PR does**:

Add an option to auto-scale on custom and external metrics.

No issue opened yet

**Test values.yaml**:

```
nginx:
  autoscaling:
    enabled: true
    customMetrics:
    - type: External
      external:
        metric:
          name: queue_messages_ready
          selector:
            matchLabels:
              queue: "worker_tasks"
        target:
          type: AverageValue
          averageValue: 30
    - type: External
      external:
        metric:
          name: queue_workers_ready
        target:
          type: AverageValue
          averageValue: 10
```

**Result:**

```
---
# Source: cortex/templates/nginx/nginx-hpa.yaml
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: release-name-cortex-nginx
  namespace: default
  labels:
    helm.sh/chart: cortex-2.1.0
    app.kubernetes.io/name: cortex
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "v1.15.1"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: nginx
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: release-name-cortex-nginx
  minReplicas: 2
  maxReplicas: 30
  metrics:
    - type: Resource
      resource:
        name: cpu
        target:
          type: Utilization
          averageUtilization: 80
    - external:
        metric:
          name: queue_messages_ready
          selector:
            matchLabels:
              queue: worker_tasks
        target:
          averageValue: 30
          type: AverageValue
      type: External
    - external:
        metric:
          name: queue_workers_ready
        target:
          averageValue: 10
          type: AverageValue
      type: External
---
```

**Checklist**


locmai:
Hi @nschad and @kd7lxl , could you take a look at this PR?

values.yaml
```diff
@@ -624,6 +660,24 @@ ruler:
...
+    # - type: External
+    #   external:
+    #     metric:
+    #       name: queue_messages_ready
+    #       selector:
+    #         matchLabels:
+    #           queue: "worker_tasks"
+    #     target:
+    #       type: AverageValue
+    #       averageValue: 30
```
nschad, line 679: shouldn't these blocks be under `customMetrics`? Looks a bit confusing with the docs or is just me?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Add option to autoscale on multiple metrics and custom metrics metrics
```
nschad, line 5:
```suggestion
* [ENHANCEMENT] Add option to autoscale on multiple metrics and custom metrics metrics #467
```

values.yaml
locmai: I believe they should be added before the block? you could view the README here - the description is on the right row. https://github.com/cortexproject/cortex-helm-chart/blob/438cea3317ade5b45b2fa7db2ea0450963438f8b/README.md

CHANGELOG.md
locmai: Updated. Thanks!

values.yaml
```diff
@@ -372,6 +372,24 @@ distributor:
...
+    customMetrics: []
```
kd7lxl, line 392: Naming it `extraMetrics` would be more inline with the naming practices in the rest of the chart.

values.yaml
kd7lxl: To get them to show up in the README, they go before the line.

The word wrapping makes the yaml example unreadable in the README. Can you experiment with ways to fix them? Maybe indenting or wrapping in <code>```</code>.

values.yaml
locmai: Thanks for the naming, I will update it in the next push

values.yaml
locmai: Yah I get them up in there. True that it is unreadable. The `basicAuthSecretName` would have the same issue.

Just tried a several way to break/escape but didn't work. Let me experiment a bit more or I could just move that out of the way as it's also not good looking in a table. For other Helm charts, they usually below the key to be like an example value for that key.

locmai: Found a way to make it looks nice on the surface README with code tag and tab way. Do you think we could go with this or just simply shorten the description is better? 

I'm happy with both.

kd7lxl: I am also happy with either solution. Formatted or below the key are both fine.

nschad (approved):

nschad:
thanks!

PR #476, opened 2023-08-07
Allow optional entry to /ruler/rule_groups endpoint

### LINKED ISSUE #475: Optionally expose Ruler rules endpoint in nginx ingress
The supplied nginx ingress does not include the Ruler rules endpoint that returns all rule groups without a customer id required.

The specific endpoint is: GET /ruler/rule_groups

Some folks may wish to keep this endpoint private because they have a public Cortex API or for something else. So, it should be optional.

AlexDHoffer:
**What this PR does**: Allow optional entry into /ruler/rule_groups endpoint via nginx proxy

Fixes #475 

**Checklist**


nschad (commented):
Custom route "injection" is already supported by setting  `.Values.nginx.config.serverSnippet`

I'm inclined to reject this PR on the basis that this would just add more spaghetti templating

AlexDCraig:
The custom route injection for this is kind of hideous:

````
nginx:
  config:
    serverSnippet:
      location ~ /ruler/rule_groups {
          proxy_pass      http://{{ template "cortex.fullname" . }}-ruler.{{ printf "%s.svc.%s:%d" .Release.Namespace .Values.clusterDomain (.Values.config.server.http_listen_port | int) }}$request_uri;
      }
````

Or is there a way to call that $rootDomain function the other routes use from the values file?

**Comment by nschad** [comment]
> The custom route injection for this is kind of hideous:
> 
> ```
> nginx:
>   config:
>     serverSnippet:
>       location ~ /ruler/rule_groups {
>           proxy_pass      http://{{ template "cortex.fullname" . }}-ruler.{{ printf "%s.svc.%s:%d" .Release.Namespace .Values.clusterDomain (.Values.config.server.http_listen_port | int) }}$request_uri;
>       }
> ```
> 
> Or is there a way to call that $rootDomain function the other routes use from the values file?

I don't know, have you tried setting $rootDomain? I assume the context might be different

File: values.yaml [hunk 1]
```diff
@@ -722,6 +722,8 @@ ruler:
       enabled: true
       readOnlyRootFilesystem: true
 
+  list_all_rule_groups_api: false
```
nschad, line 725: Can we move this under `nginx.config`? Considering that the docs say it's not part of the ruler api, this should be a bit better suited there. Also might I suggest a rename to something like `exposeRulerRulesAPIEndpoint` ?

Whatever you choose please no `snake_case`. You can also add a doc string

nschad (commented):
To be honest I can't really follow my reasoning before as to why this not worth doing.

We can definitely get this in.

PR #487, opened 2023-12-18
fix podTargetLabels in all Prometheus ServiceMonitor CRDs

nschad:
**What this PR does**:

I think we all messed-up a bit in PR #439. I randomly found this. Previously all podTargetLabels are the same as in the `purger`. Must have been a copy pasta error.

CC:
@jmeza-xyz 
@kd7lxl 

Fixes #<issue number>

**Checklist**


kd7lxl (approved):
Nice find.

jmeza-xyz:
Oh wow, apologies for the copy pasta error on my part. Thanks for catching and fixing it.

Issue #420, opened 2022-12-15
Exposing 9094 tcp port for alertmanager


AlexandreRoux:
My understanding is that alertmanager is using port 9094 to communicate between peers in cluster mode but [master/templates/alertmanager](https://github.com/cortexproject/cortex-helm-chart/tree/master/templates/alertmanager) does not have references of exposing this port.

Configuring alertmanager in cluster can be done via :
```
config
   alertmanager:
      cluster:
        peers: 'cortex-alertmanager-headless.cortex.svc.cluster.local.:9094'        
```

The above configuration seems to be working :
```
$ k get services -n cortex | grep alertmanager
cortex-alertmanager              ClusterIP   10.xx.xx.201   <none>        8080/TCP   77d
cortex-alertmanager-headless     ClusterIP   None            <none>        8080/TCP   9d

$ k describe pods/cortex-alertmanager-0 -n cortex
    Ports:         8080/TCP, 7946/TCP

$ k describe statefulset/cortex-alertmanager -n cortex
    Ports:       8080/TCP, 7946/TCP
    Host Ports:  0/TCP, 0/TCP

$ kubectl exec -ti cortex-alertmanager-0 -c alertmanager -n cortex -- /bin/sh
/ # nc -zv 127.0.0.1:9094
127.0.0.1:9094 (127.0.0.1:9094) open
/ # nc -zv cortex-alertmanager-headless.cortex.svc.cluster.local:9094
cortex-alertmanager-headless.cortex.svc.cluster.local:9094 (10.xx.xx.119:9094) open

$ k logs -f -n cortex -l app.kubernetes.io/component=alertmanager -c alertmanager
level=debug ts=2022-12-13T07:26:33.456676073Z caller=cluster.go:337 component=cluster memberlist="2022/12/13 07:26:33 [DEBUG] memberlist: Initiating push/pull sync with: 01GKWRxxxxxxxxxQSDT73 10.xx.xx.223:9094\n"
```

However I believe we should expose correctly 9094, the same as it can be done via cortex-jsonnet (https://github.com/cortexproject/cortex-jsonnet) :
```
cortex-jsonnet/manifests ∙ grep -r "9094" ./ 
.//apps-v1.StatefulSet-alertmanager.yaml:        - --alertmanager.cluster.listen-address=[$(POD_IP)]:9094
[...]

PR #494, opened 2024-04-03
fix: change headless service to gRPC and expose 9094 TCP

locmai:
## What this PR does

Follow-up the work of @AlexandreRoux from #435 and related to #420 to run the helm-docs for the README.md

> This PR corrects the alertmanager headless service with exposing gRPC port instead of HTTP.

I tested the Helm chart and I can't reproduce the EOF issue.


templates/alertmanager/alertmanager-svc-headless.yaml
```diff
@@ -17,11 +16,18 @@ spec:
...
+      name: grpc
+      targetPort: grpc
+    - port: 9094
+      protocol: UDP
+      name: alert-clu-udp
+      targetPort: alert-clu-udp
+    - port: 9094
+      protocol: TCP
+      name: alert-clu-tcp
+      targetPort: alert-clu-tcp
```
nschad, line 30: We should somehow notify the user that if the service already exists, that this will not work properly due to https://github.com/kubernetes/kubernetes/issues/39188

templates/alertmanager/alertmanager-svc-headless.yaml
```diff
@@ -17,11 +16,18 @@ spec:
   clusterIP: None
   publishNotReadyAddresses: true
   ports:
-    - port: {{ .Values.config.server.http_listen_port }}
+    - port: {{ .Values.config.server.grpc_listen_port }}
       protocol: TCP
-      name: http-metrics
-      targetPort: http-metrics
```
nschad, line 23: we should inform the user about this breaking chance in CHANGELOG.md

templates/alertmanager/alertmanager-svc-headless.yaml
```diff
@@ -1,5 +1,4 @@
 {{- if .Values.alertmanager.enabled -}}
-{{- if .Values.alertmanager.statefulSet.enabled -}}
```
nschad, line 2: Why was this removed?

locmai: I think that was Alex's change respond to your previous comment about the deployment without the headless service

> Also currently if you set statefulSet to false (the default) and deploy cortex then you only get the non-headless service of alertmanager. Makes no sense.

So he reverted the removal of the dep.yaml and also had this in the next commit.

templates/alertmanager/alertmanager-svc-headless.yaml
locmai: I have updated the CHANGELOG.md to make this clearer. Or do you think we should add a better indication that this may break if users were using the `http-metrics` port for their own usage? Similar to the old PR we have here https://github.com/cortexproject/cortex-helm-chart/pull/169

And by breaking - do you mean we should make it to v3 (too much I think?) or a line in the CHANGELOG should be good enough?

templates/alertmanager/alertmanager-svc-headless.yaml
locmai: Changelog updated.

templates/alertmanager/alertmanager-svc-headless.yaml
nschad: Yep I think this should warrant a v3.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master / unreleased
 
+* [CHANGE] Replace `http-metrics` port with `grpc` port #494
```
nschad, line 5:
```suggestion
* [CHANGE] Alertmanager: Replace `http-metrics` port with `grpc` port #494 
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master / unreleased
 
+* [CHANGE] Replace `http-metrics` port with `grpc` port #494 
+* [CHANGE] Expose 9094 TCP and UDP for gossip cluster - #494
```
nschad, line 6:
```suggestion
* [CHANGE] Alertmanager: Expose 9094 TCP and UDP for gossip cluster - #494
```

nschad (approved):

templates/alertmanager/alertmanager-svc-headless.yaml
kd7lxl: I would just not touch the `http-metrics` port, so that there is no breaking change. Adding gRPC has nothing to do with the `http-metrics` port.

locmai: That sounds good. Let me update that.

locmai: Done!

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master / unreleased
 
+* [CHANGE] Alertmanager: Add `grpc` port #494 
+* [CHANGE] Alertmanager: Expose 9094 TCP and UDP for gossip cluster #494
```
kd7lxl, line 6: These should be `ENHANCEMENT`.

kd7lxl (approved):

PR #502, opened 2024-07-01
remove default livenessProbe for compactor and storage-gateway

### LINKED ISSUE #484: Disable livenessprobe by default
I have run all cortex components for years without liveness probes.
they are bad for store-gateways because it causes re-sharding if they are under pressure.
Restarting store-gateways can take a long time to recover, same for others components.
We should disable all of them by default and let users enable them as needed.

Similar to
https://github.com/cortexproject/cortex-helm-chart/pull/263

nschad:
**What this PR does**:

Removes the default livenessProbe for compactor and storage-gateway. You can still specify a `livenessProbe` if this is desired.

As discussed in #484

Fixes #484

**Checklist**

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -1410,11 +1410,7 @@ store_gateway:
       path: /ready
       port: http-metrics
       scheme: HTTP
-  livenessProbe:
-    httpGet:
-      path: /ready
-      port: http-metrics
-      scheme: HTTP
+  livenessProbe: []
   readinessProbe:
     httpGet:
       path: /ready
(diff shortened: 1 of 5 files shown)


values.yaml
```diff
@@ -1410,11 +1410,7 @@ store_gateway:
       path: /ready
       port: http-metrics
       scheme: HTTP
-  livenessProbe:
-    httpGet:
-      path: /ready
-      port: http-metrics
-      scheme: HTTP
+  livenessProbe: []
```
kd7lxl, line 1413:
```suggestion
  livenessProbe: {}
```

values.yaml
```diff
@@ -1530,11 +1526,7 @@ compactor:
...
```
kd7lxl, line 1529:
```suggestion
  livenessProbe: {}
```

nschad: Ty!

friedrichg (approved):
thank you!

PR #506, opened 2024-08-26
Add the possibility to configure the upstream protocol in the nginx configuration for the cortex upstream

finn-evers:
**What this PR does**:
Add the possibility to configure the protocol for the upstream URLs in the nginx configuration.

When preparing to upgrade to TLS communication we ran into the problem that the nginx upstream URLs are hard-coded to http.

**Checklist**

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -1196,6 +1196,8 @@ nginx:
     mainSnippet: ""
     # -- arbitrary snippet to inject in the server { } section of the nginx config
     serverSnippet: ""
+    # -- protocol for the communication with the upstream
+    upstream_protocol: http
     setHeaders: {}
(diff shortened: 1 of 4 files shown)


nschad (approved):

values.yaml
```diff
@@ -1196,6 +1196,8 @@ nginx:
     mainSnippet: ""
     # -- arbitrary snippet to inject in the server { } section of the nginx config
     serverSnippet: ""
+    # -- protocol for the communication with the upstream
+    upstream_protocol: http
```
kd7lxl, line 1200: I was going to suggest to make it camelCase to match the others, but then I noticed there is already a mix of camelCase and snake_case 🤦‍♂️

kd7lxl (approved):

PR #513, opened 2025-01-14
Set align_queries_with_step to false as default

### LINKED ISSUE #514: Query range align_queries_with_step default value issues
Hello,

While working on a Cortex deployment I noticed that the default value in the chart for `align_queries_with_step` is **true**, although in the [Cortex official docs](https://cortexmetrics.io/docs/configuration/configuration-file/#query_range_config) and [Cortex implementation](https://github.com/cortexproject/cortex/blob/5dd107290f615239d3538ecfdfd63781fc476c31/pkg/querier/tripperware/queryrange/query_range_middlewares.go#L53) is **false**.
`f.BoolVar(&cfg.AlignQueriesWithStep, "querier.align-querier-with-step", false, "Mutate incoming queries to align their start and end with their step.")`

Because of this discrepancy and because the deployment config relied on the official docs, there were many confusions on the query range endpoint especially when working with timezone.

Another thing which was confusing was that when calling the **_query-frontend_** the query result was like `align_queries_with_step` was **true**, but when calling directly the **_querier_** the response was like `align_queries_with_step` was set to **false**. I noticed that the `queryRangeMiddlewares` are sent to `QueryFrontendTripperware` but I am not sure this is why the deployment was acting like that. Could you please confirm that this was causing the query_range differenced between querier and query-frontend?

eduardscaueru:
**What this PR does**:
Set align_queries_with_step to false as default, since the 'querier.align-querier-with-step' is false in the [official Cortex docs](https://cortexmetrics.io/docs/configuration/configuration-file/#query_range_config:~:text=%5Balign_queries_with_step%3A%20%3Cboolean%3E%20%7C%20default%20%3D%20false%5D) and [Cortex implementation](https://github.com/cortexproject/cortex/blob/5dd107290f615239d3538ecfdfd63781fc476c31/pkg/querier/tripperware/queryrange/query_range_middlewares.go#L53).

Fixes #514

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Updated default `align_queries_with_step` to **false** to match documentation #513
 * [ENHANCEMENT] Add `nginx.config.upstream_protocol` field to configure the upstream protocol in the nginx configuration #506
 * [BUGFIX] fix: upstream_protocol reference in auth_orgs #509
 * [DEPENDENCY] Update quay.io/cortexproject/cortex Docker tag to v1.18.1 #510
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * [ENHANCEMENT] Add `nginx.config.upstream_protocol` field to configure the upstream protocol in the nginx configuration #506
 * [BUGFIX] fix: upstream_protocol reference in auth_orgs #509
 * [DEPENDENCY] Update quay.io/cortexproject/cortex Docker tag to v1.18.1 #510
+* [ENHANCEMENT] Updated default `align_queries_with_step` to **false** to match documentation #513
```
nschad, line 8:
```suggestion
* [CHANGE] Updated default `align_queries_with_step` to **false** to match documentation #513
```

nschad (changes requested):

CHANGELOG.md
nschad: Also could you be so kind and fix the order in the CHANGELOG?

> the order of entries should be [CHANGE], [FEATURE], [ENHANCEMENT], [BUGFIX], [DEPENDENCY]

eduardscaueru: Yes, sorry about that.

nschad (approved):

nschad:
FYI: Looks like our github-actions are too old, I've updated them here: https://github.com/cortexproject/cortex-helm-chart/pull/515

Waiting for review from Tom

PR #517, opened 2025-01-15
Add ability to configure StatefulSet PVC retention policy

### LINKED ISSUE #516: Allow for configuration of the `persistentVolumeClaimRetentionPolicy` field on deployed StatefulSets
Our deployment of Cortex uses autoscaling for some of the components (such as ingesters). When the StatefulSet replicas scale down, we are left with dangling PVCs which we would prefer to delete. Kubernetes [supports this](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention) vial the `persistentVolumeClaimRetentionPolicy` field.

sam-mcbr:
**What this PR does**:
Adds values allowing users to configure the `persistentVolumeClaimRetentionPolicy` on StatefulSets deployed by the chart.

Fixes #516

**Checklist**

### DIFF
--- a/templates/alertmanager/alertmanager-statefulset.yaml
+++ b/templates/alertmanager/alertmanager-statefulset.yaml
@@ -21,6 +21,12 @@ spec:
     {{- toYaml .Values.alertmanager.statefulStrategy | nindent 4 }}
   serviceName: {{ template "cortex.fullname" . }}-alertmanager-headless
   {{- if .Values.alertmanager.persistentVolume.enabled }}
+  {{- if semverCompare ">=1.23-0" .Capabilities.KubeVersion.Version }}
+  {{- with .Values.alertmanager.persistentVolume.retentionPolicy }}
+  persistentVolumeClaimRetentionPolicy:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- end }}
   volumeClaimTemplates:
     - metadata:
         name: storage
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [CHANGE] Updated default `align_queries_with_step` to **false** to match documentation #513
 * [ENHANCEMENT] Add `nginx.config.upstream_protocol` field to configure the upstream protocol in the nginx configuration #506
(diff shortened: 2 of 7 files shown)


templates/alertmanager/alertmanager-statefulset.yaml
```diff
@@ -21,6 +21,11 @@ spec:
     {{- toYaml .Values.alertmanager.statefulStrategy | nindent 4 }}
   serviceName: {{ template "cortex.fullname" . }}-alertmanager-headless
   {{- if .Values.alertmanager.persistentVolume.enabled }}
+  {{- if and (semverCompare ">=1.23-0" .Capabilities.KubeVersion.Version) (.Values.alertmanager.persistentVolume.enableRetentionPolicy) }}
+  persistentVolumeClaimRetentionPolicy:
+    whenDeleted: {{ .Values.alertmanager.persistentVolume.whenDeleted }}
+    whenScaled: {{ .Values.alertmanager.persistentVolume.whenScaled }}
+  {{- end }}
```
kd7lxl, line 28: Do we get any benefit from having the bool `enableRetentionPolicy` to enable this? To me it seems simpler to just use the values if they are defined, but I get that this is a bit of a style thing.

It could be like this:
```suggestion
  {{- if (semverCompare ">=1.23-0" .Capabilities.KubeVersion.Version) }}
  {{- with .Values.alertmanager.persistentVolume.retentionPolicy }}
  persistentVolumeClaimRetentionPolicy:
    {{- toYaml . | nindent 4}}
  {{- end }}
  {{- end }}
```
The values.yaml would be like:
```yaml
alertmanager:
  persistentVolume:
    retentionPolicy: {}
    # whenDeleted: Retain
    # whenScaled: Retain
```

nschad: I'm fine with both.

sam-mcbr: @kd7lxl I like that suggestion, I'll go ahead and make the changes

kd7lxl (approved):

nschad (approved):

Issue #473, opened 2023-07-20
Create a production ready values files


friedrichg:
Most users using the helm chart have to specify a few things in the values.yaml to get their cortex to production ready status.
We should create an alternate values file to make this path easier, maybe copying some of the values specified in cortex-jsonnet.

jessequinn:
@friedrichg possibly you can start by providing what you believe is a good `values.yaml` so others can evaluate.

abctaylor:
Chiming in to say this would be incredibly helpful, especially for folks new to Kubernetes and Helm

danfinn:
As a new user working through the helm install I fully agree, there is quite a bit missing to get this installed and production ready.  I'm currently working through issues with my distributor and ingester pods OOM'ing and it looks like it's because there are no memory limits set on the pods and also GOMEMLIMIT is not configured.

danfinn:
here is the ansible task we are using to do the helm install of cortex, values are specified here:

```
- name: Helm Cortex for Prometheus
  kubernetes.core.helm:
    name: cortex
    binary_path: "{{ helm310_binary_path }}"
    kubeconfig: "{{ context_file }}"
    context: "{{ aks_name }}"
    chart_ref: cortex-helm/cortex
    chart_version: v2.2.0
    wait: true
    wait_timeout: 600s
    release_namespace: "{{ namespace }}"
    values:
      ingress:
        enabled: true
        ingressClass:
          name: "nginx"
        annotations:
          cert-manager.io/cluster-issuer: "{{ namespace }}-letsencrypt-issuer"
          kubernetes.io/ingress.class: internal-nginx
        hosts:
          - host: "cortex.{{ dns_zone }}"
            paths:
              - /
        tls:
          - hosts:
              - "cortex.{{ dns_zone }}"
            secretName: cortex-tls
      ingester:
        nodeSelector:
          agentpool: "{{ agent_pool }}"
        replicas: 4
        resources:
          limits:
            memory: "16Gi"
        env:
          - name: GOMEMLIMIT
            value: 14000MiB
      distributor:
        nodeSelector:
          agentpool: "{{ agent_pool }}"
        replicas: 4
        resources:
          limits:
            memory: "8Gi"
        env:
          - name: GOMEMLIMIT
            value: 7000MiB
      alertmanager:
        enabled: false
      ruler:
        enabled: false
      query_frontend:
        nodeSelector:
          agentpool: "{{ agent_pool }}"
      querier:
[...]

PR #542, opened 2025-06-13
Remove startup probe from compactor

friedrichg:
In some circumstances the compactor might take a long time to start and that is not really a bug. For example when there is a high number of partial blocks to read.

**What this PR does**:  Remove startup probe from compactor

**Which issue(s) this PR fixes**: it works towards #473

**Checklist**
- [X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`, `[DEPENDENCY]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
+* [CHANGE] Remove startup probes from compactors from default values. It's not recommended in general
+
 ## 2.6.0 / 2025-05-07
 
 * [DEPENDENCY] Update Helm release memcached to v6.14.0 #528
--- a/values.yaml
+++ b/values.yaml
@@ -1,4 +1,3 @@
-image:
   repository: quay.io/cortexproject/cortex
   # -- Allows you to override the cortex version in this chart. Use at your own risk.
   tag: ""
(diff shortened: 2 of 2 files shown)


kd7lxl:
> In some circumstances the compactor might take a long time to start and that is not really a bug

The startup probe you removed declares that 30 minutes is allowed for the startup phase. Can you describe in more detail why you prefer to immediately jump to the readiness probe stage?

friedrichg:
@kd7lxl I had a case where the compactors were taking over 30 minutes to start due to the number of partial blocks in S3. Shuffle sharding in compactors creates temporary partial blocks—that’s just how it works. For unrelated reasons, some compactors hadn't run for a long time, leading to a huge backlog of partial blocks. The startup probe kept killing the compactors, which in turn increased the backlog even more.

This might only be relevant for large clusters, but I’ve run clusters of all sizes without startup probes on compactors for years without any issues. I’ve only recently started using the Helm chart heavily. Until now, I’ve primarily used the cortex-sonnet repo, which doesn’t use startup probes—so that’s where my suggestion to change the default behavior is coming from

kd7lxl (approved):

friedrichg:
Yes, you are right. I had liveness probe as well. The combination of the 2 was killing compactors after 30 minutes. We don't need the startup probe alone or with liveness. Let's discourage it.

alanprot (approved):

nschad:
needs a quick rebase, but lgtm

PR #578, opened 2025-12-09
Automate release workflow

kd7lxl:
**What this PR does**:
Automates the release workflow. It is manually triggered (for now) from Actions tab. There is a helper `make bump-version` that will look at the new entries in the changelog and apply semver rules to bump the version in Chart.yaml and CHANGELOG.md. Next it packages the chart, updates the index, and opens a PR with the changes.

**Checklist**
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`, `[DEPENDENCY]`

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -21,7 +21,7 @@ jobs:
       # yamllint (https://github.com/adrienverge/yamllint) which require Python
       - uses: actions/setup-python@v6
         with:
-          python-version: '3.13'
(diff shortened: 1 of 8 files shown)


nschad:
This looks like the current "process" but honestly we should probably (if we're gonna switch). Just use github chart-releaser action. https://github.com/helm/chart-releaser

PR #594, opened 2026-02-20
Bug/ruler use wrong alertmanager url

### LINKED ISSUE #593: Ruler is using the wrongly generated Alertmanager URL
The Ruler will use the wrong hostname for AlertManager when

* Release name is not cortex 
* config.ruler.alertmanager_url is not configured
* config.ruler.enable_alertmanager_discovery is true

This happens because ruler deployment file is using "cortex.name" from template which will always resolved to "cortex-headless". However, Alertmanager headless service uses "cortex.fullname" to generate the service name which can be different from "cortex-headless".

Because of this, when the above condition is met, ruler will be deployed using the wrong alertmanager URL

amirulali:
**What this PR does**:
Use the same method as alertmanager-headless-service to generate the hostname for alertmanager_url in ruler deployment.

Fixes #593

**Checklist**
- [X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`, `[DEPENDENCY]`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -72,3 +72,4 @@ jobs:
               --body "Automated release PR for version ${{ steps.version.outputs.new_version }}" \
               --base ${{ github.ref_name }} \
               --head release-${{ steps.version.outputs.new_version }}
+          fi


nschad:
Looks alright can you sign the DCO? @amirulali

PR #609, opened 2026-03-27
fix(volumeClaimTemplates): add apiVersion and kind to volumeClaimTemplates

### LINKED ISSUE #608: ArgoCD detects constant drift on StatefulSet volumeClaimTemplate
## Bug Description

ArgoCD continuously detects drift and attempts to sync StatefulSet resources (alertmanager, compactor, ingester, store-gateway) due to missing `apiVersion` and `kind` fields in `volumeClaimTemplates`.

## Problem

When deploying the Cortex Helm chart, Kubernetes automatically populates the `apiVersion: v1` and `kind: PersistentVolumeClaim` fields in the StatefulSet's `volumeClaimTemplates` during resource creation. However, the Helm templates don't include these fields, causing ArgoCD to detect a constant difference between:

- **Desired state** (from Helm): `volumeClaimTemplates` without `apiVersion`/`kind`
- **Actual state** (in cluster): `volumeClaimTemplates` with `apiVersion`/`kind`

This results in ArgoCD sync flapping, constantly trying to "fix" a difference that reappears immediately.

## Expected Behavior

StatefulSet templates should explicitly include `apiVersion` and `kind` fields in `volumeClaimTemplates` to match the actual cluster state and prevent ArgoCD drift detection.

## Affected Components

- `templates/alertmanager/alertmanager-statefulset.yaml`
- `templates/compactor/compactor-statefulset.yaml`
- `templates/ingester/ingester-statefulset.yaml`
- `templates/store-gateway/store-gateway-statefulset.yaml`

## Reference

The Grafana Loki Helm chart uses this pattern successfully: https://github.com/grafana/loki/blob/main/production/helm/loki/templates/ingester/statefulset-ingester.yaml#L83-L85

gianvetter:
Add apiVersion and kind to volumeClaimTemplates

**What this PR does**:

Adds explicit `apiVersion: v1` and `kind: PersistentVolumeClaim` fields to `volumeClaimTemplates` in all StatefulSet templates (alertmanager, compactor, ingester, store-gateway).

While these fields are optional in Kubernetes API, Kubernetes automatically populates them when creating StatefulSets. This caused ArgoCD to constantly detect drift between the desired state (without these fields) and the actual cluster state (with these fields), resulting in continuous sync flapping.

Fixes #608

**Checklist**

### DIFF
--- a/templates/alertmanager/alertmanager-statefulset.yaml
+++ b/templates/alertmanager/alertmanager-statefulset.yaml
@@ -28,7 +28,9 @@ spec:
   {{- end }}
   {{- end }}
   volumeClaimTemplates:
-    - metadata:
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
         name: storage
         {{- if .Values.alertmanager.persistentVolume.annotations }}
         annotations:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master / unreleased
 
-* [BUGFIX] Add `apiVersion` and `kind` fields to `volumeClaimTemplates` in StatefulSets to prevent ArgoCD sync flapping #<PR_NUMBER>
+* [BUGFIX] Add `apiVersion` and `kind` fields to `volumeClaimTemplates` in StatefulSets to prevent ArgoCD sync flapping #608
 
 ## 3.2.0 / 2026-02-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
+* [BUGFIX] Add `apiVersion` and `kind` fields to `volumeClaimTemplates` in StatefulSets to prevent ArgoCD sync flapping #<PR_NUMBER>
+
 ## 3.2.0 / 2026-02-26
 
 * [ENHANCEMENT] enable readiness probe on kiwigrid/k8s-sidecar #597
--- a/templates/compactor/compactor-statefulset.yaml
+++ b/templates/compactor/compactor-statefulset.yaml
@@ -25,7 +25,9 @@ spec:
   {{- end }}
   {{- end }}
   volumeClaimTemplates:
-    - metadata:
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
         name: storage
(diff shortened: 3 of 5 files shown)


templates/alertmanager/alertmanager-statefulset.yaml
```diff
@@ -28,7 +28,9 @@ spec:
   {{- end }}
   {{- end }}
   volumeClaimTemplates:
-    - metadata:
+    - apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
```
kd7lxl, line 33: The official docs do not include the `apiVersion` and `kind` fields in the `volumeClaimTemplate`: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components
```yaml
  volumeClaimTemplates:
  - metadata:
      name: www
```

The drift you are seeing is a bug in ArgoCD, not in this chart.

kd7lxl (changes requested):

templates/alertmanager/alertmanager-statefulset.yaml
gianvetter: You're absolutely correct that it's technically an ArgoCD bug.

However, this issue has been open for 3 years (https://github.com/argoproj/argo-cd/issues/16707), and many widely-used charts have already added this workaround to prevent constant drift warnings for ArgoCD users.

For example:
- bitnami/postgresql: https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/primary/statefulset.yaml#L687
- prometheus-community/prometheus: https://github.com/prometheus-community/helm-charts/blob/250617de25fd73935709bd22366dfa87beb9737d/charts/prometheus/templates/deploy.yaml#L464

I'd prefer to align this chart with how other charts handle it to improve the experience for ArgoCD users. What do you think?

If you don't think this change is necessary, I can also set up an ignore diff in ArgoCD. That works for me too.

kd7lxl: I'd prefer to keep usage here inline with the official docs. I welcome other opinions on this topic if anyone else wants to voice them. In the short term, I recommend adding the ignore diff workaround or deploying a fork of this chart.

nschad: I don't really have a strong opionion for or against it. We would just add something redundant, right? can't think of any major disadvantages

PR #632, opened 2026-05-20
Add zone awareness support for ingester

### LINKED ISSUE #203: Support zone awareness
To support zone awareness we need to create different ingesters configurations as many zones required. Currently this is not possible through the helm install.

SungJin1212:
This pull request introduces support for zone-aware ingester deployments in the Helm chart, allowing ingesters to be distributed across multiple availability zones (zones A, B, and C).

Fixes #203 

**Checklist**


nschad:
I appreciate this but I don't really like this static config way. For example what if a customer has more than three zones? (for whatever reason)

What do you think @kd7lxl

This feels like just a whole lot of duped configs with a whole lot of pitfalls.

kd7lxl (commented):
Have you tried the technique in https://github.com/cortexproject/cortex-helm-chart/issues/203#issuecomment-1051641083

SungJin1212:
@nschad @kd7lxl 
Instead of hardcoding zone-a/b/c as separate keys, why don't you use a list like this?
Adding a 4th or 5th zone requires only a values.yaml change, no chart updates
```
zoneAwareness:
  zones:
    - name: zone-a
      replicas: 1
      nodeSelector: {}
    - name: zone-b
      replicas: 1
      nodeSelector: {}
    - name: zone-c
      replicas: 1
      nodeSelector: {}
```

SungJin1212:
@kd7lxl @nschad 
I changed it to the list way. Could you take a look when you have some time?

nschad:
We're currently looking at #668 internally. I think this has the most amount of de-dupe while also following the grafana/mirmir example.

PR #650, opened 2026-07-03
fix: avoid rendering empty env key in querier deployment

archy-rock3t-cloud:
**What this PR does**:

The querier deployment renders an empty `env:` key when no environment variables are configured. Wrapped in conditional to match existig pattern.

**Checklist**

### DIFF
--- a/templates/querier/querier-dep.yaml
+++ b/templates/querier/querier-dep.yaml
@@ -88,10 +88,10 @@ spec:
           {{- if .Values.querier.containerSecurityContext.enabled }}
           securityContext: {{- omit .Values.querier.containerSecurityContext "enabled" | toYaml | nindent 12 }}
           {{- end }}
+          {{- if .Values.querier.env }}
           env:
-            {{- if .Values.querier.env }}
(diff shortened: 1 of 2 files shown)


nschad (approved):

templates/querier/querier-dep.yaml
```diff
@@ -88,10 +88,10 @@ spec:
           {{- if .Values.querier.containerSecurityContext.enabled }}
           securityContext: {{- omit .Values.querier.containerSecurityContext "enabled" | toYaml | nindent 12 }}
           {{- end }}
+          {{- if .Values.querier.env }}
           env:
-            {{- if .Values.querier.env }}
-              {{- toYaml .Values.querier.env | nindent 12 }}
-            {{- end }}
+            {{- toYaml .Values.querier.env | nindent 12 }}
```
kd7lxl, line 93: Cleaner syntax for this is:
```suggestion
          {{- with .Values.querier.env }}
env:
            {{- toYaml . | nindent 12 }}
```

kd7lxl (approved):

PR #668, opened 2026-09-10
Add support for zone-aware replication

### LINKED ISSUE #203: Support zone awareness

timonegk:
**What this PR does**:
Add support for zone-aware replication to ingesters. Similar to #632, but with less duplication.
I have documented the migration process and tested it in a cluster with six ingesters.

Fixes #203.

**Checklist**


kd7lxl (commented):
Why would you modify the chart when the chart already supports [another technique for zone awareness](https://github.com/cortexproject/cortex-helm-chart/issues/203#issuecomment-1051641083) that doesn't require further modification? Is there something I'm missing?

nschad:
> another technique for zone awareness

That requires the admission controller, otherwise you won't get the env-vars. If that is something you are ok with running, then yes you are correct. Otherwise there is currently no true "stand-a-lone" way of deploying zone-aware ingesters

right?

kd7lxl:
> > another technique for zone awareness
> 
> That requires the admission controller, otherwise you won't get the env-vars. If that is something you are ok with running, then yes you are correct. Otherwise there is currently no true "stand-a-lone" way of deploying zone-aware ingesters
> 
> right?

Yes, I would run the admission controller. There is no need for the admission controller to be deployed in the same chart as cortex (nor would I couple them), so this is possible today. A guide doc may be the only contribution needed.

A strong benefit of the admission controller is that is does not require prior knowledge of the available zones. In contrast, the configuration method requires first collecting information about the target cluster and what zones are available, then configuring cortex for those zones. This is significant increased deployment complexity.

nschad:
> > > another technique for zone awareness
> > 
> > 
> > That requires the admission controller, otherwise you won't get the env-vars. If that is something you are ok with running, then yes you are correct. Otherwise there is currently no true "stand-a-lone" way of deploying zone-aware ingesters
> > right?
> 
> Yes, I would run the admission controller. There is no need for the admission controller to be deployed in the same chart as cortex (nor would I couple them), so this is possible today. A guide doc may be the only contribution needed.
> 
> A strong benefit of the admission controller is that is does not require prior knowledge of the available zones. In contrast, the configuration method requires first collecting information about the target cluster and what zones are available, then configuring cortex for those zones. This is significant increased deployment complexity.

on the other hand, this PR allows you have to one Deployment/StatefulSet per zone which can be beneficial if you want to do per-zone rollouts. For example facialited by the[ grafana rollout operator](https://github.com/grafana/rollout-operator)

I think there is a case to be made for both. Also the admission controller (even though the code is not complicated) is not maintained.

timonegk:
@kd7lxl we are mostly benefitting from this change because of faster rollouts. With zone-aware ingesters, you can roll out one zone at a time, so a rollout out is O(zones) instead of O(ingesters). For us, that brings rollout times down from several hours to about 15 minutes.
If I understand your proposal correctly, this is not something we could achieve with the admission controller.

kd7lxl:
> @kd7lxl we are mostly benefitting from this change because of faster rollouts. With zone-aware ingesters, you can roll out one zone at a time, so a rollout out is O(zones) instead of O(ingesters). For us, that brings rollout times down from several hours to about 15 minutes. If I understand your proposal correctly, this is not something we could achieve with the admission controller.

That is a valuable benefit!

templates/distributor/distributor-dep.yaml
```diff
@@ -54,6 +54,15 @@ spec:
           args:
             - "-target=distributor"
             - "-config.file=/etc/cortex/cortex.yaml"
+          {{- /* enable zone-awareness if it's enabled and not in migration or in migration with writePath enabled */ -}}
+          {{- if and .Values.ingester.zoneAwareReplication.enabled
+                     (or (not .Values.ingester.zoneAwareReplication.migration.enabled)
+                         .Values.ingester.zoneAwareReplication.migration.writePath) }}
+            - "-distributor.zone-awareness-enabled"
+          {{- if .Values.ingester.zoneAwareReplication.migration.enabled }}
+            - "-distributor.excluded-zones=zone-default"
```
nschad: what's that? `zone-default` ?

timonegk: It's part of the migration process. When the write path is in migration (`migration.writePath=true`), the default zone should be excluded so that we only write to the new zones.
But the zone name should be `default`, not `zone-default`. I fixed it, but that means I have to test again.

timonegk: With the wrong zone name the test worked because there were still 2/3 replicas in the other ingester zones. I will test again tomorrow with the fixed value.

timonegk: @nschad I have tested it and the default zone ingesters are now handled correctly (`cortex_ingester_active_series` goes to 0 for them once the write path is migrated).

nschad:
@kd7lxl can you have a look when you have time?

I'm biased :D 

PR should be ready.

kd7lxl (approved):

docs/guides/migrate_to_zone_aware_replication.markdown
```diff
@@ -0,0 +1,87 @@
...
+The general migration process is the following: New stateful sets are created, the write traffic is routed to them, the read traffic is routed to them, the old stateful set is disabled.
+During the migration, it is ensured that at most one ingester is unavailable at the time, and that an ingester's data is always written to persistent storage before it is shut down.
+
+The chart makes use of the [rollout-operator](https://github.com/grafana/rollout-operator) to coordinate rollouts of the stateful sets. This will automatically set the stateful set's update strategy to `OnDelete`.
+
+## Prerequisites
+
+Make sure to set the following settings before starting the migration:
+- Ingesters are deployed as a stateful set (deployment is currently not supported)
+- Autoscaling is disabled (autoscaling is not supported for zone-aware ingesters)
```
kd7lxl, line 32: While I have you here, what is the scaling strategy with zone-aware ingesters?

timonegk: I would recommend against autoscaling ingesters (zone-aware or not) and prefer static setup or manual scaling. The ingester ring is not very forgiving to changes.
I can think of the following issues:
When scaling down, you must ensure that `unregister-on-shutdown` is set to avoid the need to manually remove ingesters from the ring. However, this should not always be set lest a rolling restart will result in many changes in the ring and data will potentially be stored at different ingesters than expected while the rollout is running. To avoid new data missing from the ingester ring, you should also set the ingesters to `READONLY` for at least `query-store-after` before removing the ingester, or set `flush-blocks-on-shutdown` and wait for `query-store-after` between scaling down `floor(RF/2)` replicas.
When scaling up, new ingesters join the ring, so you should disable shuffle sharding and use `shard-by-all-labels` to avoid that only a subring of ingesters is queried (which could only contain the new ingesters), or wait for `query-store-after` between scaling up `RF-1` replicas.
The ingester HPA default values currently scales one replica per 3h, so even for the default settings, users would need to adapt it.
In a zone-aware setup, most of these concerns remain. Scaling within a zone is easier, so adding/removing multiple replicas in the same zone could be done in parallel, but the same constraints as in the single-zone setup apply between the zones (e.g. wait `query-store-after` before starting the scaledown in the second zone). I am not aware that this coordination is currently possible in Kubernetes.

nschad (approved):

